Issue #13, opened 2016-05-24
Replace categorical with discrete in Primitive


adscib:
Currently one constructor for `Primitive` is `Categorical :: [(a,LogFloat)] -> Primitive a` with arbitrary type `a`. This is a little inconvenient, and forces us to make certain restrictions on `a` which do not make sense in all contexts. It would make more sense to replace `Categorical` with `Discrete :: [LogFloat] -> Primitive Int` and then provide default implementation for `categorical` in terms of `discrete`.

PR #20, opened 2016-06-10
Resample-move SMC

yfcai:
Preliminary prototype of resample-move SMC according to Adam Scibior's idea of using a list of continuations as the state of Metropolis transition.

Changes
- Replaced module `Trace` to use a list of continuations as the state of the Markov chain, implemented via a coroutine with suspension points at primitive distributions. Alternative states (i. e., types in the class `RandomDB`) are deleted.
- Added `Inference.smcWithResampler`, the common template between `smc` and `smcFast`, implemented according to the paper draft.
- Added `Inference.smcrm` according to the paper draft. The only subtle point is that `mhStep` is specialized to `Population m` to discard new weight after transition. Only then is the posterior preserved.


adscib:
While you're at it, could you take care of #13? I think it only affects the part of the code that is to do with Trace MH and should simplify it a little.

adscib:
I just added some notes on Trace to the draft in section 2.7. Could you check how this compares to your new implementation of the Trace transformer? In particular, can we have an analogue of `mhstep` in implementation? I'm not sure if it's possible to do things this way, but it would be very convenient if that was the case.

yfcai:
Thanks for the pointer, I could work on #13 after the resample-move prototype is done.

About §2.7: I will try this approach. Right now I can't make `Tr` a monad and the work-around is less appealing. It involves posterior-preserving natural transformations into functors that are not monads.

adscib:
Is there a simple explanation for why `Tr` could not be a monad?

yfcai:
`Tr` saves a list of continuations from each primitive together with an answer. It is not clear how to combine the continuation lists sensibly in monadic bind.

adscib:
Can't we extend the continuations on the left with the RHS of `>>=`, leave the continuations on the right alone, and concatenate the two lists?

yfcai:
Each continuation takes the value of one primitive variable and re-executes the program from that point onward, reusing previous samples whenever permissible. If continuations on LHS are to be combined with `f`, they must also retain the exact primitive samples during the run that generated the RHS.

Maybe I'm not defining `T` and `Tr` correctly. I will think about it more.

yfcai:
`smcrm` is now almost exactly as it appears in the paper draft. The monad problem is solved by converting between `m (Tr a)` and coroutines with suspension points at primitive distributions. The latter is clearly a monad. The conversions are `Trace.mhState` and `Trace.mhReset`.

yfcai:
Build 11.6 failed because Travis couldn't locate `ghc`. Reason for failure seems to be the extra library dependency on `mmorph`. I will remove that.

adscib:
This is looking good - I'll take a more careful look later.

The problem with Travis is the script that caches Stack files across multiple builds. It usually works fine, but every now and then it can't find GHC. I don't think it is `mmorph` that's causing problems, but let's not add it to dependencies just yet.

adscib:
Could you explain why `mhStep'` needs to be specialized to `Particle m`? In particular, why is there no `mhStep' :: (MonadDist m) => Trace' m a -> Trace' m a`? My intuition is that this would be sufficient for SMC-RM. To be precise, from the look of it seems like `mhStep' (Trace' m) = Trace' (mhStep m)` would do the job.

Also, why is `Trace` not an instance of `MonadTrans`?

yfcai:
> why `mhStep'` needs to be specialized to `Particle m`?

The main reason is that we want to discard the weight produced by MH transition and restore the old weight. Otherwise MH transition does not preserve posterior. Here is a proof sketch.

Let `P` be the Metropolis kernel. Let `π` be the posterior distribution. `P` preserves `π` in the sense that

```
π(dy) = P(x,dy) π(dx) .
```

Let `ρ` be the prior and let `f` be the scoring function; then

```
π(dx) = f(x) ρ(dx) .
```

Substituting the 2nd equation into the 1st yields

```
π(dy) = f(x) P(x, dy) ρ(dx) .
```

Note that after the MH transition, we are sampling directly from `P(x, dy) ρ(dx)`. To retain the posterior `π(dy)`, the correct score is the old score `f(x)`, not the new one `f(y)`. Simply calling `mhStep` inside `Trace'` would use the new weight `f(y)` and would not be correct.

There could be a way to implement `mhStep'` without mentioning `Population`, probably along the line of

```
mhStep' :: (MonadBayes m) => Trace' m a -> Trace' m a  .
```

We need `MonadBayes` to be able to adjust the score after MH transition. So far I haven't managed to restore the old weight correctly. I will give it more attempts.

> why is `Trace` not an instance of `MonadTrans`?

It easily could be. I will add the `MonadTrans` instance declaration.

adscib:
I don't see how `mhStep` modifies the weight. There are two things we might call "weight".

One is the weight associated with a particle, which is kept in the `Population` monad. It is only modified when `factor` is called in `Trace'`, which only happens during the first execution. It is not modified when `factor` is called in `Trace`, which happens in `mhStep` and `mhStep'`. So this weight is not modified by the MH transition and it correctly remains the old weight.

The other weight is that associated with a trace that's called `mhPosteriorWeight` and is kept in the `Trace` monad. This weight is modified by `mhStep`, but that is the correct behaviour. `mhPosteriorWeight` in only used for computing acceptance ratio, so it should be the new weight. `marginal` doesn't reinsert `mhPosteriorWeight` into the transformed monad, which is again correct, so there is no need for it to be in any way related to the old weight.

In more practical terms, I believe your current `mhStep'` would always produce `newWeight = 1`, so there is no harm in returning the product `oldWeight*newWeight` instead of `oldWeight`. And this is exactly what would happen if you simply lifted `mhStep` to `Trace'`.

adscib:
> I will add the MonadTrans instance declaration.

Please also write `MonadBayes Trace'` instance as `factor k = Trace' $ do {lift (factor k); factor k}` if this is indeed correct.

yfcai:
Your `MonadBayes Trace'` instance seems correct; it preserves the posterior for the sprinkler model.

Inside `Trace'`, `mhStep` modifies the weight of the particle due to resuming a continuation. `Trace'` has to pass each call of `factor` to the underlying monad `Population m` so that particles may have different weights. But then the continuations inside each `MHState` carries those calls to `factor` inside `Population m`. After `mhStep`, the weight of the particle is modified by factors encountered after the first resampled primitive distribution.

adscib:
I see. That is indeed a problem and I don't have a solution for it. But it would be great to have one.

yfcai:
I've managed to make `mhStep'` generic (610fa6f317f1c49cd7d66b7b6b83e034eeff3605). The trick is to add a `WeightRecorderT` layer to save all the factors passed to the underlying monad, and later pass reciprocals of factors we want to revoke.

```
newtype Trace' m a = Trace' { runTrace' :: Trace (WeightRecorderT m) a }
```

adscib:
Nicely done!

PR #45, opened 2018-09-04
fix compile errors in stack bench

tdoris:


adscib:
Thanks for your contribution!

PR #48, opened 2018-11-23
GHC 8.6.2 compatibility: defined/applied appropriate MonadFail instances

0xCM:


adscib:
Thank you for your contribution. The MonadFail requirement is introduced by an unfailable pattern match on a pair, so instead of introducing a constraint I would prefer to modify the pattern match to make it unnecessary. If you'd like to make these changes I'll be happy to accept your pull request.

baskaransri:
The failing patterns seem to be replaceable by Data.Vector's iterateNM + toList (or just having mh and co return m (Vector a)). I'd be happy to do so if I'm not stepping on any toes.

Does the order of the trace results matter? iterateNM would provide them in reverse order to what you have now.

YPares:
Hi! Monad-bayes cannot be built on current nixpkgs and stack LTS due to this issue. @adscib I'd be happy to provide a fix

PR #50, opened 2018-12-04
Merge importance sampling from develop

esennesh:
Hi Adam,

This feature branch implements model/guide-style importance sampling for the different Traced monads. I'm going to use my fork for my own project, but I figured I should give you a pull request in case you happen to like the idea.  I've included a couple of bugfixes to my original.

Cheers,
Eli

### DIFF
--- a/monad-bayes.cabal
+++ b/monad-bayes.cabal
@@ -49,6 +49,7 @@ executable example
 library
   hs-source-dirs:      src
   exposed-modules:     Control.Monad.Bayes.Class
+                     , Control.Monad.Bayes.Distributions
                      , Control.Monad.Bayes.Free
                      , Control.Monad.Bayes.Sampler
                      , Control.Monad.Bayes.Weighted
--- a/src/Control/Monad/Bayes/Distributions.hs
+++ b/src/Control/Monad/Bayes/Distributions.hs
@@ -1,0 +1,111 @@
+module Control.Monad.Bayes.Distributions (
+  PrimitiveDistr, sample, observe,
+  uniformDist, normalDist, gammaDist, betaDist,
+  bernoulliDist, binomialDist, categoricalDist, geometricDist, poissonDist,
+) where
+
+import Control.Monad (when)
+import Control.Monad.Bayes.Class
+import Data.List
+import Data.Maybe
+import qualified Data.Vector as V
+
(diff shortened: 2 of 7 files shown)


adscib:
Yes, I think we need something like that. My only concern is that with the current unstructured representation of traces the variables are matched by order of appearance so they need to be sampled in the same order in both the model and the guide. Is there a good use case for that? I imagine in most cases we'd like to sample the variables in the guide in roughly reverse order.

esennesh:
That's a good question.  I think variational guides are a pretty good use-case there, in that, at least IME, those usually sample in the same order as the model.  Rescoring under the model then provides an ELBO.

But monad-bayes doesn't implement variational inference, so maybe it's unnecessary to build up a foundation for it ahead of time.

esennesh:
That's a good question.  I think variational guides are a pretty good use-case there, in that, at least IME, those usually sample in the same order as the model.  Rescoring under the model then provides an ELBO.

But monad-bayes doesn't implement variational inference, so maybe it's unnecessary to build up a foundation for it ahead of time.

reubenharry:
@esennesh Do you have a good reference I could check out to understand what this code is doing? (Or just a quick explanation). No worries if not - I can probably work it out from the code.

reubenharry:
Also @esennesh if you're a good person to advise on adding VI to monad-bayes, let me know :)

esennesh:
Hi Reuben,

The basic problem for adding VI to monad-bayes is going to end up being getting gradients from somewhere.  You'd need to do the whole dance of picking an AD library, getting it to behave well when composed with the monad-transformer stack in monad-bayes, and then actually doing optimization loops.

Probably the best reference for the idea of "guide programs" as proposals is Pyro, which I think coined the term.

reubenharry:
Thanks! I saw an old branch of monad-bayes (the autodiff branch) had used the ad library to do gradients, which encouraged me that it was at least conceptually possible. It wasn't entirely clear to me whether it was working, but as I'm planning to spend a few months working on monad-bayes (as a fellowship project with Tweag), I'd thought it might be interesting to try to write reference implementations (i.e. without worrying about speed) for HMC and/or ADVI. I might need guidance on that front though, so will possibly reach out to you or other people offline

PR #53, opened 2019-03-27
Fix missing parentheses in MH, causing accept ratio to be too high

jkoppel:


adscib:
Thanks. In the meantime it turned out that lengths were also the wrong way around. Hopefully all is good now.

PR #54, opened 2019-09-16
Making some pattern-matchings irrefutable so we can build on recent GHC

YPares:
This fixes https://github.com/adscib/monad-bayes/pull/48 without the need for MonadFail constraints


freckletonj:
can confirm, this is all that's needed to bring `monad-bayes` into the 21st century ;)

curiousleo:
Note that this fixes the same compilation issue as #48, #55 and #56, but this PR contains the smallest diff.

adscib:
Thank you for your contribution!

PR #56, opened 2019-10-14
fixed errors that blocked stack build for later lts

yiblet:
#### The Problem
In the latest stack lts, `x:xs <- f ...` is a compile time error if the underlying monad does not implement monad fail. Issue seemed to be that later versions of ghc require your monad to a MonadFail if you pattern match a bind on a sum type

#### The Fix
You can avoid the compiler's complaints by avoiding pattern matching.


curiousleo:
Thanks @yiblet! https://github.com/tweag/monad-bayes/pull/54 has now been merged, which fixes the same issue. `monad-bayes` now builds with recent GHC versions. I'll close this PR since the underlying issue has been resolved.

PR #58, opened 2020-01-23
expose the Trace.Common module

MMesch:
This PR exposes the `Traced.Commons` module with some useful functions and the `Trace` type:

https://github.com/adscib/monad-bayes/blob/master/src/Control/Monad/Bayes/Traced/Common.hs#L12-L20

Is there a reason that it is currently hidden?


curiousleo:
This PR cannot be merged due to a merge conflict in `monad-bayes.cabal`. Would you mind rebasing the branch?

curiousleo:
As this is a change to the public API, a bullet point under "Unreleased" in `CHANGELOG.md` would be in order.

PR #63, opened 2020-02-17
Use Buildkite for CI

mrkkrp:


stack.yaml
```diff
@@ -1,6 +0,0 @@
-flags: {}
```
curiousleo, line 1: Let's keep `stack.yaml` for local development.

nix/nixpkgs/default.nix
```diff
@@ -0,0 +1,9 @@
+let
+  rev = "19.09";
+  sha256 = "0mhqhq21y5vrr1f30qd2bvydv4bbbslvyzclhw0kdxmkgg3z4c92";
+  nixpkgs = builtins.fetchTarball {
+    url = "https://github.com/NixOS/nixpkgs/archive/${rev}.tar.gz";
+    inherit sha256;
+  };
+  pkgs = import nixpkgs { config.allowUnfree = true; };
```
curiousleo: Is `allowUnfree` necessary?

mrkkrp: In this case perhaps not.

monad-bayes.cabal
```diff
@@ -33,6 +37,13 @@ executable example
                      , HMM
                      , LDA
                      , Dice
+  if flag(dev)
+    ghc-options: -Wall -Werror -Wcompat
```
curiousleo:
```suggestion
    ghc-options: -Wall -Wcompat
```

monad-bayes.cabal
```diff
@@ -67,7 +78,13 @@ library
                      , statistics >= 0.14.0 && < 0.16
                      , transformers >= 0.5.2 && < 0.6
                      , vector >= 0.12.0 && < 0.13
-  ghc-options:         -Wall -fno-warn-redundant-constraints
+  if flag(dev)
+    ghc-options: -Wall -Werror -Wcompat
```
curiousleo:
```suggestion
    ghc-options: -Wall -Wcompat
```

monad-bayes.cabal
```diff
@@ -95,7 +112,13 @@ test-suite monad-bayes-test
                      , transformers
                      , log-domain
                      , vector
-  ghc-options:         -Wall -threaded -rtsopts "-with-rtsopts=-N -M1g -K1g"
+  if flag(dev)
+    ghc-options: -Wall -Werror -Wcompat
```
curiousleo:
```suggestion
    ghc-options: -Wall -Wcompat
```

monad-bayes.cabal
```diff
@@ -128,7 +151,13 @@ benchmark speed-bench
                      , mwc-random
                      , containers
                      , process
-  ghc-options:         -rtsopts "-with-rtsopts=-M1g -K1g"
+  if flag(dev)
+    ghc-options: -Wall -Werror -Wcompat
```
curiousleo:
```suggestion
    ghc-options: -Wall -Wcompat
```

curiousleo (commented):
Let's remove `-Werror` for now, fix the warnings in a separate PR, and then add `-Werror` to ensure we don't introduce warnings again: https://github.com/tweag/monad-bayes/issues/66

stack-ghc822.yaml
```diff
@@ -1,6 +1,3 @@
-flags: {}
+resolver: lts-11.22
```
curiousleo:
```suggestion
resolver: lts-11.0
```
The point here is that we would like to test all changes for their backwards-compatibility. As such the most useful LTS version to test against is the initial release. The idea here is that if something builds against `lts-11.0`, it is guaranteed (?) to work with `lts-11.X` for any valid `X`. On the other hand, it is possible to have something successfully build against `lts-11.22` but not against `lts-11.0` (e.g. if there is a library `lib` in both `lts-11.0` and in `lts-11.22`, but the version of `lib` included in `lts-11.22` has additional functions - then if we use those additional functions, we'll be able to build successfully against `lts-11.22` but fail against `lts-11.0`).

stack-ghc844.yaml
```diff
@@ -0,0 +1,3 @@
+resolver: lts-12.26
```
curiousleo:
```suggestion
resolver: lts-12.0
```

stack-ghc865.yaml
```diff
@@ -0,0 +1,3 @@
+resolver: lts-14.27
```
curiousleo:
```suggestion
resolver: lts-14.0
```

default.nix
```diff
@@ -0,0 +1,60 @@
...
+  #       (super.callCabal2nix "monad-bayes" source { }) "dev";
+  #   };
+  # };
+  # haskellPackagesGhc84 = mkHaskellPackages "ghc844";
+  # haskellPackagesGhc86 = mkHaskellPackages "ghc865";
+  # haskellPackagesGhc88 = mkHaskellPackages "ghc881";
+  # haskellPackages = haskellPackagesGhc86;
+  monad-bayes-ghc82 = mkMonadBayes "stack-ghc822.yaml";
+  monad-bayes-ghc84 = mkMonadBayes "stack-ghc844.yaml";
+  monad-bayes-ghc86 = mkMonadBayes "stack-ghc865.yaml";
```
curiousleo, line 19: Does `haskell-nix` work with GHC 8.8? If so, it would be good to build against 8.8 too.

mrkkrp: Yes, but I'm not sure `monad-bayes` works with 8.8 right now. I did not want to have to fix it because it is outside of the scope for this PR.

stack-ghc822.yaml
mrkkrp: Well, as you wish, but I do not find this reasoning particularly convincing. It is extremely rare for a package to e.g. work with `lts-11.X` and stop working for `lts-11.(X+1)`. Most likely you are just choosing more buggy snapshot of the ecosystem.

default.nix
curiousleo: I am sure `monad-bayes` works with 8.8 right now. I've had to test this for the Stackage PR: https://github.com/commercialhaskell/stackage/pull/5189. So any failure that occurs here is due to the BuildKite integration and not due to the library being incompatible with GHC 8.8.

stack-ghc822.yaml
curiousleo: >  It is extremely rare for a package to e.g. work with `lts-11.X` and stop working for `lts-11.(X+1)`.

That's exactly my point - this is why I think it makes sense to build against `lts-11.0`. At least if packages adhere to the PVP, anything that builds against `lts-11.0` will, by induction, build against `lts-11.X` for any `X`.

default.nix
curiousleo: Thanks!

stack-ghc881.yaml
```diff
@@ -0,0 +1,3 @@
+resolver: nightly-2020-01-21
```
curiousleo: Could you add a comment here describing why `lts-15.0` doesn't work? I think that as soon as circumstances change, we'll want to switch to `lts-15.0` and a comment describing the current issue will be useful then.

stack-ghc844.yaml
```diff
@@ -0,0 +1,3 @@
+resolver: lts-12.15 # first lts-12.X which uses ghc844
```
curiousleo, line 1: IIRC we had to use `ghc844` because `ghc843` was not provided by the pinned version of nixpkgs. Is this still relevant now we use the haskellNix overlay? I.e. could this be `lts-12.0` again now?

mrkkrp: Yes it is still relevant. Apparently `haskell.nix` picks up compilers from what is already defined in nixpkgs.

curiousleo: > Apparently haskell.nix picks up compilers from what is already defined in nixpkgs.

I started `nix-shell` and right now it is compiling GHC, so I don't think GHC is being picked up from nixpkgs - otherwise it would just download the built version from Hydra, no?

mrkkrp: I do not have the answer. I only know that if `ghc843` is not present in nixpkgs it fails.

curiousleo (commented):
Thanks for the effort, Mark. I just hope this isn't going to be too much of a burden to maintain.

**Advantages:**
- Once the caches are hot, the BuildKite builds are really fast.

**Disadvantages:**
- The way this is set up, we can only use GHC versions that are in nixpkgs, and only Stackage snapshots built with GHC versions that are in nixpkgs
- This requires _building_ GHC and the base libraries to open `nix-shell`, which takes a very long time (it's been running for a couple of hours now)

I'll merge this as soon as I've managed to build once.

PR #82, opened 2020-03-04
Build benchmarks on CI

### LINKED ISSUE #77: Build benchmarks in CI
Benchmarks are not currently built in CI, but they should be.

`ssm-bench` broke due to recent refactorings. This breakage was not caught in CI:
```console
$ stack --stack-yaml stack-ghc881.yaml build --bench --no-run-benchmarks
[...]
Building benchmark 'ssm-bench' for monad-bayes-0.1.0.0..
[1 of 2] Compiling NonlinearSSM
            
/home/leo/Code/monad-bayes/models/NonlinearSSM.hs:16:36: error:
    • Variable not in scope: sq :: Double -> Double
    • Perhaps you meant ‘seq’ (imported from Prelude)
   |        
16 | mean x n = 0.5 * x + 25 * x / (1 + sq x) + 8 * cos (1.2 * fromIntegral n)
   |                                    ^^
            
            
--  While building package monad-bayes-0.1.0.0 using:
      /home/leo/.stack/setup-exe-cache/x86_64-linux-nix/Cabal-simple_mPHDZzAJ_3.0.0.0_ghc-8.8.1 --builddir=.stack-work/dist/x86_64-linux-nix/Cabal-3.0.0.0 build lib:monad-bayes exe:example bench:speed-bench bench:ssm-bench --ghc-options " -fdiagnostics-color=always"
    Process exited with code: ExitFailure 1
```

mrkkrp:
Close #77.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -11,13 +11,27 @@ let
     "^stack.*\.yaml$"
     "^test.*$"
   ];
-  mkMonadBayes = stackYaml: pkgs.haskell-nix.stackProject {
-    inherit stackYaml;
-    src = pkgs.lib.sourceByRegex ./. source;
-  };
-  monad-bayes-ghc84 = mkMonadBayes "stack-ghc844.yaml";
-  monad-bayes-ghc86 = mkMonadBayes "stack-ghc865.yaml";
-  monad-bayes-ghc88 = mkMonadBayes "stack-ghc881.yaml";
+  mkMonadBayes = stackYaml: doBench:
+    let
+      project = pkgs.haskell-nix.stackProject {
+        inherit stackYaml;
+        src = pkgs.lib.sourceByRegex ./. source;
+      };
+      benchmarks = project.monad-bayes.components.benchmarks;
+    in
+      pkgs.recurseIntoAttrs {
+        "lib-and-test" = project.monad-bayes;
+      } // (
+        if doBench
+        then {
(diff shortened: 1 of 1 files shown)


default.nix
```diff
@@ -11,13 +11,27 @@ let
...
     "^test.*$"
   ];
-  mkMonadBayes = stackYaml: pkgs.haskell-nix.stackProject {
-    inherit stackYaml;
-    src = pkgs.lib.sourceByRegex ./. source;
-  };
-  monad-bayes-ghc84 = mkMonadBayes "stack-ghc844.yaml";
-  monad-bayes-ghc86 = mkMonadBayes "stack-ghc865.yaml";
-  monad-bayes-ghc88 = mkMonadBayes "stack-ghc881.yaml";
+  mkMonadBayes = stackYaml: doBench:
```
curiousleo, line 14: Nit: `doBench` suggests that the benchmarks are run, but this only builds them, right? How about `buildBench` instead?

curiousleo (approved):

default.nix
mrkkrp: This is just the conventional name for this. It's called `doBenchmark` in vanilla nix I think. It is silly to actually run benchmarks in nix build anyway.

curiousleo: Cool, I rectract my nit then :)

PR #91, opened 2020-03-11
Run tests on CI

mrkkrp:


curiousleo (approved):
Thanks! I think at some point we should take another look at https://github.com/input-output-hk/haskell.nix/issues/470 to figure out how this is meant to be done, but this fixes the issue at hand so I'm happy.

mrkkrp:
There appears to be a problem with GHC 8.4 though. Generation of Haddocks fails. No, I do not know why adding tests to our build triggers this. I'll take another look tomorrow.

mrkkrp:
So, apparently GHC 8.4 parser chokes on LaTeX input. The question is why we did not see it previously. I disabled running tests with that version of the compiler, which makes the build green again, not sure if we can do better. This is pretty strange.

curiousleo:
The failure comes from Haddock choking on `src/Control/Monad/Bayes/Class.hs`, which is part of the library, but that only happens when we compile the tests.
So it seems like Haddock is run on the library iff the tests are compiled. Strange.

https://buildkite.com/tweag-1/monad-bayes/builds/95#661fa01e-4ef1-4c07-ad71-c9e8cebdf762/16-149

I'd like to build as much as possible on as many GHC versions as possible. Could you perhaps separate out the benchmark compilation from the test compilation and only disable test compilation on GHC 8.4? It seems like the benchmarks build just fine with GHC 8.4, so let's keep that.

mrkkrp:
You think we added the additional argument to `mkMonadBayes` without good reason? I think there was a problem with benchmarks which prevented us from building them with GHC 8.4.4.

mrkkrp:
Here is why we do not build benchmarks with GHC 8.4.4: https://buildkite.com/tweag-1/monad-bayes/builds/67#5e4c4450-9adf-48c6-bb9f-0ea3fde503e1

curiousleo:
> Here is why we do not build benchmarks with GHC 8.4.4: https://buildkite.com/tweag-1/monad-bayes/builds/67#5e4c4450-9adf-48c6-bb9f-0ea3fde503e1

Sorry, I have such a short memory!

PR #94, opened 2020-04-14
(admin) use GitHub project template

curiousleo:
From github.com/tweag/project.


curiousleo:
No need - there is now an organisation-wide template in place: https://github.com/tweag/.github

PR #102, opened 2020-06-09
add proposal distribution to mh sampler

MMesch:
So far this is pure refactor. We can expand on this to add proposal distributions.

### DIFF
--- a/src/Control/Monad/Bayes/Traced/Common.hs
+++ b/src/Control/Monad/Bayes/Traced/Common.hs
@@ -65,20 +65,31 @@ bind dx f = do
   t2 <- f (output t1)
   return $ t2 {variables = variables t1 ++ variables t2, density = density t1 * density t2}
 
--- | A single Metropolis-corrected transition of single-site Trace MCMC.
-mhTrans :: MonadSample m => Weighted (FreeSampler m) a -> Trace a -> m (Trace a)
-mhTrans m t@Trace {variables = us, density = p} = do
+-- | a proposal distribution over randomness
+singleVariableProposal :: MonadSample m => [Double] -> m [Double]
+singleVariableProposal us = do
   let n = length us
-  us' <- do
-    i <- discrete $ discreteUniformAB 0 (n -1)
-    u' <- random
-    case splitAt i us of
-      (xs, _ : ys) -> return $ xs ++ (u' : ys)
-      _ -> error "impossible"
+  i <- discrete $ discreteUniformAB 0 (n -1)
+  u' <- random
+  case splitAt i us of
+    (xs, _ : ys) -> return $ xs ++ (u' : ys)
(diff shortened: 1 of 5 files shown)


src/Control/Monad/Bayes/Traced/Common.hs
```diff
@@ -65,21 +65,32 @@ bind dx f = do
...
+  let n = length us
+  i <- discrete $ discreteUniformAB 0 (n -1)
+  u' <- random
+  case splitAt i us of
+    (xs, _ : ys) -> return $ xs ++ (u' : ys)
+    _ -> error "impossible"
+
+-- | A single Metropolis-corrected transition of single-site Trace MCMC
+-- | with a proposal distribution p([x_j]^i+1 | [x_j]^i). The x_j correspond
+-- | to the randomness of the sample, defined over the domain [0, 1].
```
curiousleo, line 95: What's `i`?

src/Control/Monad/Bayes/Traced/Common.hs
```diff
@@ -65,21 +65,32 @@ bind dx f = do
   t2 <- f (output t1)
   return $ t2 {variables = variables t1 ++ variables t2, density = density t1 * density t2}
 
--- | A single Metropolis-corrected transition of single-site Trace MCMC.
-mhTrans :: MonadSample m => Weighted (FreeSampler m) a -> Trace a -> m (Trace a)
-mhTrans m t@Trace {variables = us, density = p} = do
+-- | a proposal distribution over randomness
```
curiousleo, line 83: IMO "a proposal distribution over randomness" is not very descriptive. I don't know what it means.

curiousleo (commented):
Thanks! The code looks good, I think the Haddocks can be more descriptive.

src/Control/Monad/Bayes/Traced/Common.hs
MMesch: the same as in the original code: the index of the random variable that get's modified in the proposal distribution to another sample from `uniform 0 1`. All other random variables are kept fix.

src/Control/Monad/Bayes/Traced/Common.hs
MMesch: there was some discussion how to call the numbers in the interval 0 1 that are the original source of a probabilistic computation. We settled calling them `randomness :: [Double]` where `Double` really is limited to the interval 0 1 which is not enforced atm. If we call this randomness _r_, a proposal distribution over randomness gives you P(r'|r), the probability to get a _r'_ for a given _r_. In other words it gives you the probability to draw a new proposed sample from a given one in the MCMC chain. In Haskell terms _P(r'|r)_ translates to `[Double] -> m [Double]` where the first argument is _r_ and the second one the conditioned distribution _P(r')_.

PR #105, opened 2020-06-26
initial simple mh test

MMesch:
This is just to get started with a little sampler test for the `mh` function.

### DIFF
--- a/test/TestTraced.hs
+++ b/test/TestTraced.hs
@@ -1,0 +1,19 @@
+{-# LANGUAGE TypeFamilies #-}
+
+module TestTraced where
+
+import Control.Monad.Bayes.Class (uniformD)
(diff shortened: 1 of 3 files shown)


test/TestTraced.hs
```diff
@@ -0,0 +1,19 @@
...
+result :: IO [Int]
+result = do
+  samples <- sampleIOfixed $ prior $ mh 10000 $ uniformD [0, 1, 2] 
+  return $ take 9000 samples
+
+check :: [Int] -> Bool
+check ls = abs (mean ls - 1) < 0.01
+
+mean :: [Int] -> Double
+mean ls = fromIntegral (Prelude.sum ls) / fromIntegral (length ls)
```
MMesch, line 19: we should probably import this from somewhere @curiousleo

reubenharry:
We now have a similar test, so closing

PR #106, opened 2020-06-27
To understand proposals

idontgetoutmuch:
@curiousleo - here's a hand-rolled MCMC for inference for conjugate priors (so it's easy to check the chains converge). My next step is to duplicate this in monad-bayes but I think @MMesch already provided such an example.

### DIFF
--- a/Alt.ipynb
+++ b/Alt.ipynb
@@ -1,0 +1,460 @@
+{
+ "cells": [
+  {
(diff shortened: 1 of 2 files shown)


idontgetoutmuch:
Now I understand a bit more about how monad-bayes works, I don't think we need this.

PR #111, opened 2021-04-29
(ci) build with GHC 8.10

curiousleo:


reubenharry:
I'll close since we now have GHC 9.2

PR #112, opened 2021-05-06
Replace mwc

idontgetoutmuch:
This an experimental branch replacing MWC with the standard RNG in random-1.2 and with random-fu for the distributions (MWC is an RNG and distributions all in one but I need multivariate normals).


idontgetoutmuch:
Closing in favour of https://github.com/tweag/monad-bayes/pull/148

PR #117, opened 2021-11-15
Remove Anglican and friends so that benchmarks run

idontgetoutmuch:


curiousleo:
Hm, looks like this did not automatically trigger a BuildKite build. I've kicked one off manually: https://buildkite.com/tweag-1/monad-bayes/builds/147

curiousleo (approved):
LGTM, thanks!

PR #122, opened 2022-01-12
many updates to build infra 

ghost:
* updated to ghc 8.8.4
* dropped ghc 8.6.5 and 8.4.4
* added ghc 8.10.7 and 9.0.2

These changes also necessitated updating haskell.nix and I took the liberty of updating various other pieces of how things are built to use more of haskell.nix and also switched to using flakes for specifying the inputs.

I verified that building with `nix-build ...` works and that we can build with `nix-shell --run 'cabal build'`.


idontgetoutmuch:
My nix fu is not good enough to comment on the nix changes. @MMesch @mrkkrp are you able to review these changes? Or suggest some who can?

mrkkrp:
I will review this later today.

.buildkite/pipeline.yml
```diff
@@ -3,16 +3,20 @@ steps:
...
+    timeout_in_minutes: 100
+  - label: Build and test on GHC 8.10
+    command: |
+      nix-build -A monad-bayes-ghc810 --no-out-link
+    timeout_in_minutes: 100
+  - label: Build and test on GHC 9
+    command: |
+      nix-build -A monad-bayes-ghc9 --no-out-link
+    timeout_in_minutes: 100
+  - label: Build a shell environment
```
mrkkrp: Why do you need to do this separately? From what I understood the shell is going to be built as part of each corresponding `monad-bayes-*` attribute?

flake.nix
```diff
@@ -0,0 +1,33 @@
+{
```
mrkkrp, line 1: I don't have experience with Nix flakes so probably someone else needs to take a look.

.buildkite/pipeline.yml
ghost: In [`mkMonadBayes` we only create a `shellFor`](https://github.com/tweag/monad-bayes/pull/122/files#diff-245563af9e34754739b12b294f3324c1d42c116ace7ce8735e21dfbc82d8e064R12) attribute which is a project function for constructing a shell environment specific to that ghc version. 

If we were to move the constructed [shell](https://github.com/tweag/monad-bayes/pull/122/files#diff-245563af9e34754739b12b294f3324c1d42c116ace7ce8735e21dfbc82d8e064R47) to be an attribute of the attribute set created by `mkMonadBayes`, then you would be absolutely correct (and, as an added bonus, that would make it easier to add a `ghc` argument to `shell.nix` to allow us to switch ghc versions easily).

mrkkrp: Then I don't understand why bind `"shellFor"` to a function in `recurseIntoAttrs`. What is it used for?

purefn: It's there to create the `shell` attribute in `default.nix`, which is used by `shell.nix`. `mkMonadBayes` returns a subset of the attributes in the haskell.nix project. 

If the question is more about, "why is the `default.nix` structured this way?" then I don't have an answer to that. I mostly tried to leave it as I found it, just updating what was happening behind the scenes. The way I saw it, I had three choices: leave the `fetch*` for nixpkgs and haskell.nix as they were or switch to some managed solution. I chose to switch to a managed solution, which means I had to choose between niv and flakes. I chose flakes because they seem the most future proof.

mrkkrp: > It's there to create the shell attribute in default.nix, which is used by shell.nix. mkMonadBayes returns a subset of the attributes in the haskell.nix project.

But `defaultProject` is defined as

```nix
defaultProject = projects.monad-bayes-ghc810;
```

So whatever is in there comes from `./nix/projects.nix` and not from an application of `mkMonadBayes`. This is what confused me.

purefn: Oooh I see now. Sorry, you are absolutely correct and `shellFor` in `mkMonadBayes` can be removed. I forgot I had refactored that to not be used any longer. I will remove it.

ghost: Oooh I see now. Sorry, you are absolutely correct and shellFor in mkMonadBayes can be removed. I forgot I had refactored that to not be used any longer. I have removed it.

reubenharry:
In the interest of getting this merged (it would be a nice thing to merge before I start trying to make more substantive changes to the code), what do people (@mrkkrp @purefn @idontgetoutmuch ) think are the blocking issues? Or who would be a good person for me to nag? :)

purefn:
@reubenharry From my perspective, I think there is really just 2 issues:

1. The build is failing because the version of nix on the CI machines is old. We can either revert to using `flake-compat` or upgrade the version of nix on the build machines. I'm fine doing either.
2. The thing that has really been holding this up, though, is getting an approval from someone with nix (and nix flake) experience.

I'd love to get this wrapped up and merged.

reubenharry:
Sounds good! Unfortunately, I don't have any nix know-how. Does @MMesch know the right person to ask for help with this? (If in the long term there is no-one to advise on how to do nix stuff, we might want to consider getting rid of nix support)

MMesch:
Yes, I can ask about Nix things.

purefn:
> we might want to consider getting rid of nix support

I don't think that would be ideal. Nix is the primary way I build things and is super useful in CI as well, since most of the build can be easily cached. One thing I would like to see changed (and asked about at one time) was if we might switch to using cachix and GitHub Actions, instead of the homegrown stuff we are using for this project now.

aspiwack:
I've got to say, I don't think buildkite has work great for us for open source projects. It sounds reasonable to move to github actions. I've just had a look at the Nix stuff. It looks a little complex for what it's trying to achieve. I assume that there are reasonable simplifications which we could apply.

idontgetoutmuch:
I agree with @purefn that we should move to cachix and GitHub actions with a simpler nix derivation. I could probably do this but I am not convinced it would be a good use of my time. I am at best average at nix and a noob at GitHub actions.

@reubenharry I hope this is not grandmothers and eggs but the advantage of this approach is that if it builds locally it builds in CI (of course this isn't true but it's almost true).

purefn:
I will work on converting over to use GitHub actions. Does Tweag already have a cachix account that I should use?

purefn:
There have been several comments saying that the nix expressions are complicated. I'd like to better understand what that means, as I see much of this as pretty standard boilerplate. The most complicated bit is configuring tools for the nix shell, and is an unfortunate necessary complexity with haskell.nix. We could decide not to provide common haskell dev tools in the shell and force people to bring their own, but that doesn't tend to work out too well with things like HLS. What bits do other people consider complicated?

aspiwack:
> I will work on converting over to use GitHub actions. Does Tweag already have a cachix account that I should use?

There is. The access token is available as a Github secret in `TWEAG_CACHIX_AUTH_TOKEN`.

aspiwack:
> There have been several comments saying that the nix expressions are complicated. I'd like to better understand what that means, as I see much of this as pretty standard boilerplate. The most complicated bit is configuring tools for the nix shell, and is an unfortunate necessary complexity with haskell.nix. We could decide not to provide common haskell dev tools in the shell and force people to bring their own, but that doesn't tend to work out too well with things like HLS. What bits do other people consider complicated?

Why are we using Haskell.nix? This sounds overkill for this project.

purefn:
> Why are we using Haskell.nix? This sounds overkill for this project.

It's simply what I found when I got here, so I went with it. I can't comment on whether it's more or less complex than the Haskell infra in nixpkgs. I haven't used it in some time because on most projects I've worked on it lacked desired properties.

purefn:
Ah now I remember why haskell.nix is a good choice. We use the stack config files to create the projects for ghc 8.8, 8.10, and 9. The nixpkgs infra doesn't provide any ability to use the stack project files.

I will try and get it to work, but trying to support three different versions of GHC with the nixpkgs haskell infra might be difficult and far more complicated.

idontgetoutmuch:
I have

```
let

myHaskellPackageOverlay = self: super: {

  myHaskellPackages = super.haskellPackages.override {
    overrides = hself: hsuper: rec {

      random-fu = self.haskell.lib.addBuildDepends (super.haskell.lib.dontHaddock (super.haskell.lib.dontCheck (
        hself.callCabal2nixWithOptions "random-fu" (builtins.fetchGit {
          url = "https://github.com/lehins/random-fu";
          rev = "1eb5c0080618a691cd9fa81ac3d0ea29edeb084f";
          ref = "switch-to-random";
        }) "--subpath random-fu" { }
      ))) [ ];

      rvar = self.haskell.lib.addBuildDepends (super.haskell.lib.dontHaddock (super.haskell.lib.dontCheck (
        hself.callCabal2nixWithOptions "rvar" (builtins.fetchGit {
          url = "https://github.com/lehins/random-fu";
          rev = "1eb5c0080618a691cd9fa81ac3d0ea29edeb084f";
          ref = "switch-to-random";
        }) "--subpath rvar" { }
      ))) [ ];

      random = self.haskell.lib.addBuildDepends (super.haskell.lib.dontHaddock (super.haskell.lib.dontCheck (
        hself.callCabal2nixWithOptions "rvar" (builtins.fetchGit {
          url = "https://github.com/haskell/random";
          rev = "edae4f7908f3c7e00be1094034a4a09cd72ab35e";
        }) "" { }
      ))) [ ];

      hashable = super.haskell.lib.doJailbreak hsuper.hashable;

      mwc-random = hself.callHackage "mwc-random" "0.15.0.1" {};

      random-fu-multivariate =  hsuper.random-fu-multivariate;

    };
  };
};

in

{ nixpkgs ? import (fetchTarball "https://github.com/NixOS/nixpkgs/archive/21.11.tar.gz")
  { config.allowBroken = true;
    overlays = [ myHaskellPackageOverlay ]; }, compiler ? "default", doBenchmark ? false }:

 let

  haskellDeps = ps: with ps; [
    base
    derive-storable free generic-deriving
    hmatrix ieee754 log-domain mtl monad-coroutine mwc-random
    random random-fu random-fu-multivariate
    statistics vector
    (nixpkgs.haskell.lib.dontCheck criterion)

    reflection ad
  ];

in

  nixpkgs.stdenv.mkDerivation {
  name = "env";
  buildInputs = with nixpkgs.rPackages; [
    nixpkgs.cabal-install
    (nixpkgs.myHaskellPackages.ghcWithPackages haskellDeps)
  ];
  }
```

and my example is

```
{-# OPTIONS_GHC -Wall #-}

module SmcExample (run, run') where

import Control.Monad.Bayes.Population
import Control.Monad.Bayes.Inference.SMC
import Control.Monad.Bayes.Sequential (Sequential)

import Numeric.Log (Log)
import Data.List
import qualified Data.Vector as V
import Data.Vector ((!))
import Data.Maybe (catMaybes)


example :: MonadInfer m => m Bool
example = do
  x <- bernoulli 0.5
  condition x
  return x

run :: IO [(Bool, Log Double)]
run = (sampleIO . runPopulation. smcSystematic 4 4) example

stratified :: MonadSample m => V.Vector Double -> m [Int]
stratified weights = do
  let bigN = V.length weights
  dithers <- V.replicateM bigN (uniform 0.0 1.0)
  let positions = V.map (/ (fromIntegral bigN)) $
                  V.zipWith (+) dithers (V.map fromIntegral $ V.fromList [0 .. bigN - 1])
      cumulativeSum = V.scanl (+) 0.0 weights
      coalg (i, j) | i < bigN =
                       if (positions!i) < (cumulativeSum!j)
                       then Just (Just j, (i + 1, j))
                       else Just (Nothing, (i, j + 1))
                   | otherwise =
                       Nothing
  return $ map (\i -> i - 1) $ catMaybes $ unfoldr coalg (0, 0)

resampleStratified ::
  (MonadSample m) =>
  Population m a ->
  Population m a
resampleStratified = resampleGeneric stratified

smcStratified ::
  MonadSample m =>
  Int ->
  Int ->
  Sequential (Population m) a ->
  Population m a
smcStratified = sir resampleStratified

run' :: IO [(Bool, Log Double)]
run' = (sampleIO . runPopulation. smcStratified 4 4) example
```

If I then `ghci -isrc` I can

```
*SmcExample> run
run
[(True,0.12500000000000003),(True,0.12500000000000003),(True,0.12500000000000003),(True,0.12500000000000003)]
*SmcExample> run'
run'
[(True,0.12500000000000003),(True,0.12500000000000003),(True,0.12500000000000003),(True,0.12500000000000003)]
```

I am only slightly above noob at nix so I probably being hopelessly naive but why does it need to be more complicated? Is it to support `stack`?

Also at this stage given the library is going to undergo quite a lot of change, do we want to support 3 versions of ghc? Sure when it's a bit more stable and our improvements are in place.

aspiwack:
> Ah now I remember why haskell.nix is a good choice. We use the stack config files to create the projects for ghc 8.8, 8.10, and 9. The nixpkgs infra doesn't provide any ability to use the stack project files.
> 
> I will try and get it to work, but trying to support three different versions of GHC with the nixpkgs haskell infra might be difficult and far more complicated.

I guess that makes sense. The story is that we use stack to build, but we have a Nix derivation for CI and stuff.

I'm always a little worried about Haskell.nix, because it seems to be a source of fragility in other projects.

As to what is actually complex, looking into the code, I see that there are several exported attributes in the `default.nix` file, and I have no idea why (here maybe just documentation would help). Another curious thing is the definition of `mkMonadBayes`, in which I truly have no idea what's going on (I don't remember what `recurseIntoAttributes` does, but it's still surprising that we need to call such a function in this context, for instance).

Most seem to just define a script. This is rather surprising as well, why not just write scripts in `.sh` files? They would be more readable (thanks to syntax highlighting), they could be shellchecked, etc… And they could also be used by non-nix users.

Even in the context of Haskell.nix, I would say that there is some fat that we can shed, here, wouldn't you say?

purefn:
I did try out a switch to the nixpkgs haskell infra, https://github.com/tweag/monad-bayes/commit/39b0591af02c644994e5667858740e73a69f280e. It may be slightly simpler, for now. Fortunately, the only dependency I had to override was for `mwc-random` to use an older version than what's in nixpkgs. If this is preferrred I'll merge that change into this branch.

> As to what is actually complex, looking into the code, I see that there are several exported attributes in the default.nix file, and I have no idea why (here maybe just documentation would help). Another curious thing is the definition of mkMonadBayes, in which I truly have no idea what's going on (I don't remember what recurseIntoAttributes does, but it's still surprising that we need to call such a function in this context, for instance).

These are more things that are existed prior to me arriving. I didn't change them - in fact, I went somewhat out of my way to keep them as they are - because I was not sure why they were the way they are and didn't want to be overly aggressive as a new guy. `recurseIntoAttributes` sets the an attribute on the attribute set so that `nix-build` will recurse into the the attribute set to look for derivations. This is common if you want to allow someone to do `nix-build -A foo.bar` to build the `bar` derivation or `nix-build -A foo` to build `bar` and all the other derivations in `foo` at the same time. I would not mind at all cutting these out completely and simplifying `default.nix`.

> Most seem to just define a script. This is rather surprising as well, why not just write scripts in .sh files? They would be more readable (thanks to syntax highlighting), they could be shellchecked, etc… And they could also be used by non-nix users.

Again, something that predates my involvement. Although, I do tend to write utility scripts like these this way. We certainly can move them over to external shell scripts and then wrap them in the nix shell to provide the dependencies. The main danger in doing that is that non-nix users could very well have different versions of the tools used in the scripts and the results would not be consistent. But I'm happy to make that change if it is something others feel strongly about.

aspiwack:
> Again, something that predates my involvement.

I'm not blaming you :slightly_smiling_face: . I'm just pointing out where things look fragile to me, and which are opportunities for improvement.

> The main danger in doing that is that non-nix users could very well have different versions of the tools used in the scripts and the results would not be consistent.

Non-nix users always run the risk of non-consistent behaviour, but that's understood. I see that the benefit derived from Nix is the ability to refer to the dependencies in the script rather than relying on a path set by nix-shell (this also avoid to have to deal with missing dependencies gracefully). But I think that Nix is not mainstream enough for us to require Nix for anybody to effectively contribute. So it's important to think of this use-case.

idontgetoutmuch:
> I did try out a switch to the nixpkgs haskell infra, [39b0591](https://github.com/tweag/monad-bayes/commit/39b0591af02c644994e5667858740e73a69f280e). It may be slightly simpler, for now. Fortunately, the only dependency I had to override was for `mwc-random` to use an older version than what's in nixpkgs. If this is preferrred I'll merge that change into this branch.

I am surprised you had to go to an old version of `mwc-random`. Is the PR available? I can take a look at why this is necessary.

purefn:
> I am surprised you had to go to an old version of `mwc-random`. Is the PR available? I can take a look at why this is necessary.

I hadn't opened a PR until just now, https://github.com/tweag/monad-bayes/pull/133

The reason I had to use an older version of `mwc-random` is due to the upper bound on it, `< 0.15`. The version in nixpkgs is 0.15.1.

purefn:
> I'm not blaming you slightly_smiling_face . I'm just pointing out where things look fragile to me, and which are opportunities for improvement.

Sorry if that came off as being defensive, that's not how I intended it. I meant it more in the way of an explanation to why I left it as-is. 

> Non-nix users always run the risk of non-consistent behaviour, but that's understood. I see that the benefit derived from Nix is the ability to refer to the dependencies in the script rather than relying on a path set by nix-shell (this also avoid to have to deal with missing dependencies gracefully). But I think that Nix is not mainstream enough for us to require Nix for anybody to effectively contribute. So it's important to think of this use-case.

I've extracted the scripts to stand alone shell script files. I've made sure that the nix shell includes all the dependencies required to run them.

purefn:
> There is. The access token is available as a Github secret in `TWEAG_CACHIX_AUTH_TOKEN`.

@aspiwack I switched over to using GitHub Actions in commit https://github.com/tweag/monad-bayes/pull/122/commits/99277822ed1d8a12a6b192504ba6bf65611c17a8#diff-9b47550554bc75b6fb07dd141b5d1c6249aca44f3753f861f551f6814347b953. But I'm getting this error when it gets to pushing to cachix.

```
cachix: FailureResponse (Request {requestPath = (BaseUrl {baseUrlScheme = Https, baseUrlHost = "cachix.org", baseUrlPort = 443, baseUrlPath = ""},"/api/v1/cache/tweag/00cggwyr939hsqw01kyn87ll5d51yq1n.narinfo"), requestQueryString = fromList [], requestBody = Just ((),application/json;charset=utf-8), requestAccept = fromList [], requestHeaders = fromList [("Authorization","<REDACTED>")]), requestHttpVersion = HTTP/1.1, requestMethod = "POST"} (Response {responseStatusCode = Status {statusCode = 401, statusMessage = "Unauthorized"}, responseHeaders = fromList [("Date","Fri, 22 Apr 2022 03:44:35 GMT"),("Content-Type","application/json;charset=utf-8"),("Transfer-Encoding","chunked"),("Connection","keep-alive"),("Cache-Control","no-cache"),("CF-Cache-Status","DYNAMIC"),("Expect-CT","max-age=604800, report-uri=\"[https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct\](https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct/)""),("Report-To","{\"endpoints\":[{\"url\":\"https:\\/\\/a.nel.cloudflare.com\\/report\\/v3?s=pOCx90h7HaQLVK9gcxSdhyTzOvxEPDr4zA6eUQdsJhuW9eSGlouXrVmkn3wD9fR%2B81NZca6HssSyN%2BWVJmmE6KpC0zUR65Gkix47Ztq1myI8JfQbLNTpkSVaiQ4%3D\"}],\"group\":\"cf-nel\",\"max_age\":604800}"),("NEL","{\"success_fraction\":0,\"report_to\":\"cf-nel\",\"max_age\":604800}"),("Server","cloudflare"),("CF-RAY","6ffb5c9d0b339645-SJC")], responseHttpVersion = HTTP/1.1, responseBody = "{\"error\":\"No narinfo signature found. Is $CACHIX_SIGNING_KEY set? To generate one, run `cachix generate-keypair`.\"}"})
```

Is there something I missed?

.github/workflows/nix.yml
```diff
@@ -0,0 +1,23 @@
...
+      - uses: cachix/install-nix-action@v16
+        with:
+          nix_path: nixpkgs=channel:nixos-unstable
+          extra_nix_config: |
+            trusted-public-keys = hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ= cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=
+            substituters = https://hydra.iohk.io https://cache.nixos.org/
+      - uses: cachix/cachix-action@v10
+        with:
+          name: tweag
+          authToken: "${{ secrets.TWEAG_CACHIX_AUTH_TOKEN }}"
```
aspiwack: @purefn Try the following (disclaimer: I don't know what I'm doing):
```suggestion
          authToken: "${{ secrets.TWEAG_CACHIX_AUTH_TOKEN }}"
          signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
```

purefn: > @purefn Try the following (disclaimer: I don't know what I'm doing):

Giving it a whirl. I just wasn't sure what the name of variable might be. I can try prefixing it with `TWEAG_` if this doesn't work.

aspiwack:
@garbas can you advise?

garbas:
@aspiwack Who is the owner of this Cachix account? What plan do we have there. This is actually a topic I want to bring up as for a general discussion at tweag: Do we:
1. spend time to setup S3 bucket and maintain it,
2. we pay Cachix to do this for us.

Generally having a global cache is not a good idea. It is rather easy to setup a per repository cache. That is also the whole idea of Cachix.

I gave also a quick look at the flake setup you have here. I would do things little differently to make it easier to maintain and be more future-flake proof. If you want I can do a review on the nix setup as well, but I definitely don't want to overstep :)

aspiwack:
@mboes is the only one who has Tweag access. (I personnally don't see a problem with having a single cache, though). This is not the right place for a larger internal discussion. For now, what I would like is for @purefn to be unblocked.

idontgetoutmuch:
@purefn and I have discussed the current CI goals and we agreed that until @reubenharry has completed their improvements, we should have something very simple (there is no point in supporting 3 versions of ghc while in such a state of flux). I think that is what the change now does but my nix / CI fu is not sufficiently high to comment on the specifics.

I do note that

1. The build still fails :-(
2. I get

```
Wandle% nix-shell shell.nix
nix-shell shell.nix
error: attribute 'aarch64-darwin' missing

       at /Users/dom/monad-bayes/shell.nix:4:1:

            3|
            4| (builtins.getFlake (toString ./.)).devShells.${system}.default
             | ^
(use '--show-trace' to show detailed location information)
```

```
Wandle% git log --oneline -1
f64190e (HEAD -> build-updates, origin/build-updates) remove unused file
```

purefn:
@idontgetoutmuch I added support for more systems in the flake, so you shouldn't have that problem if you pull and try again.

I also disabled cachix for now in the build. It would be nice to have to speed up the builds, but the builds only take 10m right now so it isn't critical.

garbas:
@purefn @idontgetoutmuch @aspiwack I took some time to look at the Nix setup. To save on time I just pushed the changes here, but I tried to contain each change in a separate commit that way if you disagree with something it should be easy to revert the change. I also tried to write somewhat descriptive commit messages, but if there is something unclear why I did something please ask away.

.envrc
```diff
@@ -0,0 +1 @@
+use flake
```
purefn, line 1: I don't think we need this.

default.nix
```diff
@@ -1,64 +1,14 @@
...
-        "shellFor" = project.shellFor;
-        "lib-and-test" = project.monad-bayes;
-      } // (
-      if doBenchAndTest
-      then {
-        "ssm-bench" = benchmarks.ssm-bench;
-        "speed-bench" = benchmarks.speed-bench;
-        "checks" = project.monad-bayes.checks;
+      fetchTarball {
+        url = "https://github.com/edolstra/flake-compat/archive/${lock.nodes.flake-compat.locked.rev}.tar.gz";
```
purefn, line 8: Why not use `builtins.getFlake`? I know it's not available in older versions of nix, but it's much nicer than using this. Are there still a significant number of people using older nix versions? Do we have stats on that?

flake.nix
```diff
@@ -0,0 +1,83 @@
...
+    ];
+    extra-trusted-public-keys = [
+      "tweag-monad-bayes.cachix.org-1:tmmTZ+WvtUMpYWD4LAkfSuNKqSuJyL3N8ZVm/qYtqdc="
+    ];
+  };
+
+  inputs = {
+    nixpkgs.url = "nixpkgs/nixos-unstable";
+    flake-compat.url = "github:edolstra/flake-compat";
+    flake-compat.flake = false;
```
purefn, line 16: Can we clean this up? i.e. 

```
flake-compat = { 
  url = ...;
  flake = ...;
}
```

The duplication hurts my eyes.

flake.nix
```diff
@@ -0,0 +1,83 @@
...
+    self,
+    nixpkgs,
+    flake-compat,
+    flake-utils,
+    pre-commit-hooks,
+  }:
+    flake-utils.lib.eachSystem
+    [
+      flake-utils.lib.system.x86_64-linux
+      flake-utils.lib.system.aarch64-linux
```
purefn, line 36: @idontgetoutmuch has an M1 MBP. With just these he won't be able to build or get into a nix shell.

.envrc
garbas: Quite few nix users use direnv (or nix-direnv) as well which makes entering development environment even nicer experience.

https://github.com/nix-community/nix-direnv

Adding a file with one line is a nice way of adding support for direnv.

default.nix
garbas: Simple use `flake-compat` to provide compatibility with older version of Nix. This is also what is tested.

Adding `flake-compat` to `flake.nix` is just a sneaky way of using `flake.lock` to pin the `flake-compat` flake input.

flake.nix
garbas: You choice :)

flake.nix
```diff
@@ -0,0 +1,83 @@
...
+        monad-bayes-dev = pkgs.mkShell {
+          inputsFrom = [monad-bayes.env];
+          shellHook =
+            pre-commit.shellHook
+            + ''
+              echo "=== monad-bayes development shell ==="
+            '';
+        };
+
+        pre-commit = pre-commit-hooks.lib.${system}.run {
```
purefn, line 73: @idontgetoutmuch do we want to prevent commits if we violate linters?

flake.nix
```diff
@@ -0,0 +1,83 @@
...
+          src = self;
+          hooks = {
+            alejandra.enable = true;
+            hlint.enable = true;
```
purefn, line 78: I have mixed feelings about hlint. It is often wrong. I'm not sure it should gate commits, but will defer to @idontgetoutmuch.

flake.nix
garbas: Yeah, the point here was to be explicit about the systems you want to support. Feel free to add more.

Ideally you would also build it on that system.

flake.nix
garbas: Up to you how you decide.

My mantra is, if you have linters (which produce stable enough output) that enforce them and stop fighting about code style :)

flake.nix
garbas: I followed what is used in the `master`, but other haskell linters are available -> https://github.com/cachix/pre-commit-hooks.nix#haskell

I assume we want to use ormolu :)

flake.nix
```diff
@@ -0,0 +1,83 @@
...
+            pre-commit.shellHook
+            + ''
+              echo "=== monad-bayes development shell ==="
+            '';
+        };
+
+        pre-commit = pre-commit-hooks.lib.${system}.run {
+          src = self;
+          hooks = {
+            alejandra.enable = true;
```
purefn, line 76: Why this instead of `nixpkgs-fmt` or another?

flake.nix
```diff
@@ -0,0 +1,83 @@
...
+          shellHook =
+            pre-commit.shellHook
+            + ''
+              echo "=== monad-bayes development shell ==="
+            '';
+        };
+
+        pre-commit = pre-commit-hooks.lib.${system}.run {
+          src = self;
+          hooks = {
```
purefn, line 75: If we want this, I would add `cabal-fmt`.

flake.nix
garbas: alejandra is currently actively developed fmt tool. It fixes a lot of problems that nixpkgs-fmt has. Most likely is alejandra is going to become a default formatter (see https://github.com/NixOS/rfcs/pull/101)

flake.nix
garbas: I would only advice to not complicate this PR and for now use existing linters (as in master branch). Then open PRs to improve on the situation.

.envrc
purefn: I think the problem is that some folks have `direnv` and want to use it for some things, but not for others. For instance, I use it for client work to bring env vars into scope. But I don't want to use it for nix shells. I strongly prefer to use `nix develop` for that. Now, whenever I run a command in my clone of this repo, I get the annoying error message that the directory is blocked.

default.nix
purefn: I understand the use. I'm just wondering if the copied/complicated code is warranted. Are there really that many nix users that haven't upgraded?

flake.nix
purefn: I would rather not use ormolu. In fact, I'd rather not use any Haskell code formatters that don't strictly stick to imports. They all exhibit pathological tendencies in even simple cases.

garbas:
@purefn in last commit I've added also the darwin system system and I'm trying to add things to also build on darwin (I will probably need few commits to get this right)

flake.nix
purefn: By that logic we should be using `nixpkgs-fmt` ;)

purefn:
@garbas Did GH add M1 as a possible runner?

garbas:
> @garbas Did GH add M1 as a possible runner?

As far am I'm aware no. But I hope to add self hosted runners for M1 in next months. And also arm linux GH runners.

garbas:
I had to also comment out cachix until I will configure billing. I'll try to do this tomorrow.

.envrc
aspiwack: I don't think the `.envrc` file should be committed in the repo either. In principle, somebody could want to use a different `.envrc` that doesn't use Nix, but another virtual environment gizmo.

garbas: The usual thing I saw to use for both cases is to source `.envrc.local` from `.envrc` (if the file exists ofcourse) then you get best of both worlds

default.nix
garbas: Yes, there are many. By default on NixOS you still get Nix 2.3. And I think even with the next NixOS release you will have users that don't upgrade.

.envrc
idontgetoutmuch: FYI I use direnv but I wouldn't expect an `.envrc` file to be in the repo but maybe that's my ignorance.

What is the difference between having `use nix` and `use flake`? I and I think @reubenharry are very naive when it comes to nix.

flake.nix
idontgetoutmuch: I would say yes - let's stick to a standard - for clarity yes we do want to prevent commits if it violates the linter.

flake.nix
idontgetoutmuch: > I have mixed feelings about hlint. It is often wrong. I'm not sure it should gate commits, but will defer to @idontgetoutmuch.

Ha - I am in danger of contradicting myself but I share your concerns about hlint.

idontgetoutmuch (commented):
@purefn @garbas thanks very much for working on this. I and I am sure @reubenharry really appreciate having a nice base from which to start our improvements.

idontgetoutmuch:
Having discussed this some more, it seems we do want `ormolu` and we should deactivate all hints from `hlint` (I don't know how to do this though). We can then add a whitelist later if we so choose.

.envrc
aspiwack: > The usual thing I saw to use for both cases is to source `.envrc.local` from `.envrc` (if the file exists ofcourse) then you get best of both worlds

I don't understand the proposed solution. Can you elaborate?

purefn:
Complete sidenote, probably better discussed elsewhere, but

> Generally having a global cache is not a good idea. It is rather easy to setup a per repository cache. That is also the whole idea of Cachix.

I couldn't disagree more. This is terrible UX for nix users, as they end up with a great many substituters if they're involved in even a handful of projects. Even with just three or four I get annoyed at having to sit and wait for nix to check all of them for every nix store path on a large project. I see one of the great weaknesses of the nix ecosystem as being a missing central cache, like repo1.maven.org. Obviously this presents some difficulties, but I don't think that a federated cache would be impossible.

> spend time to setup S3 bucket and maintain it,

I don't think we should do this. Previous experience doing this tells me it is not as easy as it at first seems. What you end up wanting is a service that sits in front of the bucket and does some accounting and validation. And you don't really want to just use `nix copy` either, because it's a bit naive, so you end up wanting to write a custom client to work with your custom cache service. Before you know it, you've reimplemented cachix.

.envrc
purefn: Ya, I don't understand why we'd have a stub `.envrc` file and a git ignored `.envrc.local` file. It's really no different than having an ignored `.envrc`, but adds a bit of indirection/complication.

.github/workflows/nix.yml
```diff
@@ -0,0 +1,42 @@
...
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+
+jobs:
+
+  lint:
```
purefn, line 12: Is there cost savings with having this run in a separate job? I understand the logic that the results will be the same no matter what system it is run on, so why not just run it once. But considering how fast the linting process is and how long it takes to start a job and have all the required dependencies pulled down, it might be faster to just run it `n` times.

purefn: On second thought, the advantage of having a separate job is when there is a failure you only see it in one place. So now I'm torn between having to wait for a job I know will pass and having that clarity when it might not.

purefn:
@idontgetoutmuch I've made the changes to add ormolu to the linters and applied the formatting suggestions it had. I disabled all the hlint suggestions and we can start whitelisting things in the future. 

I've also gone ahead and made sure that all the linting tools are in the dev shell.

idontgetoutmuch (approved):
I'd love to merge this but I don't know why buildkite is failing.

PR #123, opened 2022-01-13
Replace mwc

ghost:
Based on the changes from this branch, https://github.com/idontgetoutmuch/monad-bayes/tree/replace-mwc, with updated builds for GHC.

I ran the benchmarks on each branch then combined the reports, [both.html.gz](https://github.com/tweag/monad-bayes/files/7866424/both.html.gz) (you'll need to `gunzip` it because github wouldn't let me attach a plain HTML page).

The results look good. There is not much different in timings between mwc and the random-fu branches.


ghost:
- [x] We're using a fork of `random-fu` that uses `random`, so we should get those changes merged into `random-fu` before merging this.

- [x] `random-fu` also doesn't yet support GHC 9. There is an open PR, https://github.com/haskell-numerics/random-fu/pull/76, to fix it. It manually expands the broken template haskell bits. We can either take that as-is or try and find another way.

ghost:
I was able to get `monad-bayes` compiling with ghc 8.8.4 and 9.0.2 (thanks to @idontgetoutmuch for fixing up `random-fu` for me!). I ran the speed benchmark with each of them (
[speed.html.gz](https://github.com/tweag/monad-bayes/files/7872958/speed.html.gz)). 

The results seem to indicate that on ghc 8.10.7, using `random-fu` with `random` gives about the same performance as using `mwc-random`. The story is different on ghc 8.8.4 and 9.0.2 - many of the benchmarks indicate that `mwc-random` is faster with those versions of ghc.

I have no explanation for why that would be at this time.

idontgetoutmuch:
Looking good :-) we should fix the build failure. I'll take a look once I am back from vacation.

ghost:
Build failures are fixed now.

idontgetoutmuch:
I think we have inadvertently restricted ourselves to pure RNGs only (entirely my fault if we have). I can't see a way of now using MWC. If I try

```
foo :: SamplerIO (Gen RealWorld) b -> IO b
foo m = do
  g <- MWC.create
  sampleIO m g
  ```

and then try 

```
*Main> :t foo (undefined :: (MonadIO m, MonadSample m) => m ())
:t foo (undefined :: (MonadIO m, MonadSample m) => m ())

<interactive>:1:6: error:
    • No instance for (RandomGen (Gen RealWorld))
        arising from an expression type signature
    • In the first argument of ‘foo’, namely
        ‘(undefined :: (MonadIO m, MonadSample m) => m ())’
      In the expression:
        foo (undefined :: (MonadIO m, MonadSample m) => m ())
```

I think we need to use this (rather than `RandomGen` - also I removed most of the methods and instances so as to make things clearer):

```
type StatefulGen :: * -> (* -> *) -> Constraint
class Monad m => StatefulGen g m where
  uniformWord64 :: g -> m GHC.Word.Word64
  	-- Defined in ‘System.Random.Internal’

instance (s ~ PrimState m, PrimMonad m) => StatefulGen (Gen s) m
  -- Defined in ‘System.Random.MWC’
```

We can always make a pure RNG into a monadic RNG.

I am pretty sure that pureMT (https://hackage.haskell.org/package/mersenne-random-pure64) will work but I suggest we really want to still be able to use MWC (albeit via `random-fu` rather than using `mwc-random`'s distributions).

Caveat: It may be that MWC can be used directly and I have convinced myself that it can't

**Review by idontgetoutmuch — CHANGES_REQUESTED** [review]
See my latest comment

File: src/Control/Monad/Bayes/Sampler.hs [hunk 1]
```diff
@@ -76,34 +88,27 @@ instance Applicative SamplerST where
 instance Monad SamplerST where
   (SamplerST x) >>= f = SamplerST $ x >>= runSamplerST . f
 
--- | Run the sampler with a supplied seed.
```
idontgetoutmuch, line 79: Once we have figured out how to use monadic generators we should put this back in (with a constraint to only allow it to be used on RNGs which support (are instances of) `FrozenGen`.

ghost:
Sounds good. I was wondering if we could generalize further. I'll work on that this week.

ghost:
@idontgetoutmuch I was able to steal away some time this week to work on this a bit more. See the latest commit, which I think is in the direction you were looking for. There is one huge caveat on this, and that is that this [instance](https://github.com/tweag/monad-bayes/pull/123/commits/bc056c7ffd290922455a4e1a3269e2e99be3db1b#diff-fb99ff4a9b407e2cd97fd82998d0c81a801b03f39a956f070465a53b89513fe7R74)

```
instance RS.StatefulGen g (ST RealWorld) => MonadSample (SamplerIO g) where
```
 requires `UndecideableInstances`. Without it, we see this error

```
src/Control/Monad/Bayes/Sampler.hs:73:10: error:
    • The constraint ‘RS.StatefulGen g (ST RealWorld)’
        is no smaller than the instance head ‘MonadSample (SamplerIO g)’
      (Use UndecidableInstances to permit this)
    • In the instance declaration for ‘MonadSample (SamplerIO g)’
   |
73 | instance RS.StatefulGen g (ST RealWorld) => MonadSample (SamplerIO g) where
   |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I'm not sure how strongly you feel about that, but I haven't been able to convince myself there is a reasonable way around it.

idontgetoutmuch:
This works :-)

```
{-# LANGUAGE ScopedTypeVariables #-}

{-# OPTIONS_GHC -Wall #-}

module Main where

import System.Random.MWC as MWC
import System.Random as RS
import System.Random.Stateful as RS

import Control.Monad.ST (RealWorld, stToIO)


-- | A toss of a six-sided die.
die :: forall m . MonadSample m => m [Int]
die = go 10 []
  where
    go :: (Num a, Enum a) => Int -> [a] -> m [a]
    go 0 xs = return xs
    go n xs = do x <- uniformD [1 .. 6]
                 go (n - 1) (x:xs)

foo :: SamplerIO (Gen RealWorld) b -> IO b
foo m = do
  g <- MWC.create
  sampleIO m g

bar :: SamplerIO (STGenM StdGen RealWorld) b -> IO b
bar m = do
  g <- stToIO (newSTGenM (RS.mkStdGen 42))
  sampleIO m g

main :: IO ()
main = do
  bar die >>= print
  foo die >>= print
```

I am a bit concerned that the user has to use `stToIO` but will think about that later.

I don't have any problem with `UndecidableInstances`. I assume it just means there is a possibility that the compiler may fail in some way.

idontgetoutmuch:
I tried

```
baz :: SamplerIO (IOGenM StdGen) b -> IO b
baz m = do
  g <- newIOGenM (RS.mkStdGen 42)
  sampleIO m g
```

but

```
*Main> baz die
baz die

<interactive>:123:5: error:
    • No instance for (MonadIO (ST RealWorld))
        arising from a use of ‘die’
    • In the first argument of ‘baz’, namely ‘die’
      In the expression: baz die
      In an equation for ‘it’: it = baz die
```

idontgetoutmuch:
I suppose I was expecting a type signature a bit more like

```
sampleJO :: StatefulGen g m => SamplerIO g a -> g -> m a
sampleJO = undefined
```

and then I can do

```
foo1 :: PrimMonad m => SamplerIO (Gen (PrimState m)) b -> m b
foo1 m = do
  g <- MWC.create
  sampleJO m g

bar1 :: SamplerIO (STGenM StdGen s) b -> ST s b
bar1 m = do
  g <- newSTGenM (RS.mkStdGen 42)
  sampleJO m g

baz1 :: MonadIO m => SamplerIO (IOGenM StdGen) b -> m b
baz1 m = do
  g <- newIOGenM (RS.mkStdGen 42)
  sampleJO m g
````

Now the `stToIO` is gone but I still get a constraint error for `baz1`

```
*Main> :t baz1 die
:t baz1 die

<interactive>:1:6: error:
    • Could not deduce (MonadIO (ST RealWorld))
        arising from a use of ‘die’
      from the context: MonadIO m
        bound by the inferred type of it :: MonadIO m => m [Int]
        at <interactive>:1:1
    • In the first argument of ‘baz1’, namely ‘die’
      In the expression: baz1 die
```

ghost:
@idontgetoutmuch Check out the latest commit. I've been toying with it a bit the last few days and so far this is the best compromise I've found that allows the example you posted. With this newer, more generic `Sampler` I'm not sure that `SamplerIO` and `SamplerST` make as much sense. 

The biggest issue I haven't been able to quite wrap my head around with this new formulation is going from a `Sampler` that uses `ST`, like `STGenM g s`, to something that uses `IO`, like `IOGenM g`. I tried a few things, but with all the universal type quantification it just doesn't quite work out.

reubenharry:
@idontgetoutmuch Thoughts on the status of this PR? Are there things I could help with here, and are we still planning to merge it?

idontgetoutmuch:
I don't think we can merge it as it is but I would like to make monad-bayes independent of the RNG.

purefn:
Ya, right now it's not ready as-is. I was hoping to get feedback from @idontgetoutmuch (https://github.com/tweag/monad-bayes/pull/123#issuecomment-1040560726) before trying to finalize any API changes. I've sort of dropped the ball on following up and seeing if he had a chance to look at the changes - I've had a lot of client work and personal life stuff happening lately.

idontgetoutmuch:
Closing in favour of https://github.com/tweag/monad-bayes/pull/148

PR #128, opened 2022-03-23
Docs, comments, and minor features

reubenharry:
Mainly, this adds documentation, generated by Sphinx. It's possible that I should be checking in less, and/or having the CI run Sphinx itself. I'm not sure. If/when this is merged, I'll use netlify to serve the docs; this is what I'm doing already, but currently I point netlify to my fork of monad-bayes.


reubenharry:
Linter throws some errors in the CI. Unsure how to do linting locally (do I need the nix setup?), so not yet sure how to fix.

idontgetoutmuch:
Should I comment on the docs somehow? For example I can see here: https://monad-bayes.netlify.app/probprog.html#sequential-monte-carlo-particle-filtering that the type signature has a typo:

```
run :: IO [Bool]
run = (sampleIO . runPopulation. smcSystematic 4 4) example
```

should be

```
run :: IO [(Bool, Log Double)]
run = (sampleIO . runPopulation. smcSystematic 4 4) example
```

reubenharry:
Yep, definitely let me know about typos, and thanks for catching that one! I mostly pushed this to try out making a pull request (because I'm not good at that sort of thing) and see if the gitlab tests would fail (they did). I expect I'll continue to add to the docs, either in this pull request or in future. 

(Note that I've marked this pull request as a draft for now.)

reubenharry:
@idontgetoutmuch @MMesch Do you know if there's anyone I can get advice from re. fixing build failures? The CI has some complaint about the Python, with some nix related error message, and it's a bit over my head, unfortunately.

idontgetoutmuch:
> @idontgetoutmuch @MMesch Do you know if there's anyone I can get advice from re. fixing build failures? The CI has some complaint about the Python, with some nix related error message, and it's a bit over my head, unfortunately.

I can take a look - and if I get stuck there are a lot of experts in Tweag

reubenharry:
@idontgetoutmuch I think I've finally got CI working here, so this should be ready to look at. As noted above, there's probably a way for the CI to build the docs automatically, but for now I'm just building the html locally and checking it into the repo.

idontgetoutmuch (changes requested):
> Monad-bayes is a library for doing probabilistic programming in Haskell, using probability monads. The emphasis is on composition of inference algorithms implemented in terms of monad transformers.

To me the benefit of using monad-bayes rather a more popular PPL like Stan is not only that I can specify my model in the language and do inference (and I hope prior and posterior predictions)  but that I can specify the inference algorithm and even create new ones. Being able to specify PMMH and SMC^2 in one line is something I have not seen in other PPLs. What do you think of modifying the opening paragraph to include something like this?

As as statistician or modeller, I think the first I would look for is how to do linear regression. Stan does this: https://mc-stan.org/docs/2_29/stan-users-guide/linear-regression.html. I did look at the pymc documentation but I wasn't easily able to follow how I was supposed to do linear regression (@simeoncarstens - what am I doing wrong with pymc?). What do you think about replacing the model in https://monad-bayes.netlify.app/probprog.html#user-guide by https://mc-stan.org/docs/2_29/stan-users-guide/linear-regression.html?

reubenharry:
> > Monad-bayes is a library for doing probabilistic programming in Haskell, using probability monads. The emphasis is on composition of inference algorithms implemented in terms of monad transformers.
> 
> To me the benefit of using monad-bayes rather a more popular PPL like Stan is not only that I can specify my model in the language and do inference (and I hope prior and posterior predictions) but that I can specify the inference algorithm and even create new ones. Being able to specify PMMH and SMC^2 in one line is something I have not seen in other PPLs. What do you think of modifying the opening paragraph to include something like this?
> 
> As as statistician or modeller, I think the first I would look for is how to do linear regression. Stan does this: https://mc-stan.org/docs/2_29/stan-users-guide/linear-regression.html. I did look at the pymc documentation but I wasn't easily able to follow how I was supposed to do linear regression (@simeoncarstens - what am I doing wrong with pymc?). What do you think about replacing the model in https://monad-bayes.netlify.app/probprog.html#user-guide by https://mc-stan.org/docs/2_29/stan-users-guide/linear-regression.html?

Those both sound like good ideas to me - I'll work on them.

reubenharry:
@idontgetoutmuch Updated re. both those suggestions. Let me know if it looks better: https://monad-bayes.netlify.app/probprog.html

docs/source/probprog.md
```diff
@@ -0,0 +1,552 @@
+# User Guide
+
+$$
+foo
```
idontgetoutmuch: Why does it say `foo` here?

docs/source/probprog.md
```diff
@@ -0,0 +1,552 @@
...
+`regression` takes observations of `xs` and `ys`, and using the prior expressed by `paramPriorRegression`, returns the posterior conditioned on the observations.
+
+This is the *model*. To perform *inference* , suppose we have a data set of `xs` and `ys` like:
+
+![](_static/priorpred.png)
+
+We could then run the model as follows:
+
+```haskell   
+mhRunsRegression = sampleIO $ prior $ mh 1000 $ regression xs ys
```
idontgetoutmuch, line 44: Where did `prior` come from? What does it do?

docs/source/probprog.md
```diff
@@ -0,0 +1,552 @@
...
+
+```haskell
+paramPriorRegression :: MonadSample m => m (Double, Double, Double)
+paramPriorRegression = do
+    slope <- normal 0 2
+    intercept <- normal 0 2
+    noise <- gamma 4 4
+    return (slope, intercept, noise)
+
+regression :: (MonadInfer m) => [Double] -> [Double] -> m (Double, Double, Double)
```
idontgetoutmuch, line 28: I got warnings

```
DocExamples.hs:13:5: warning: [-Wunused-matches]
    Defined but not used: ‘params’
   |
13 |     params@(slope, intercept, noise) <- paramPriorRegression
   |     ^^^^^^

DocExamples.hs:14:5: warning: [-Wunused-do-bind]
    A do-notation statement discarded a result of type ‘[()]’
    Suppress this warning by saying
      ‘_ <- forM
              (zip xs ys)
              (\ (x, y)
                 -> factor $ normalPdf (slope * x + intercept) (sqrt noise) y)’
   |
14 |     forM (zip xs ys) (\(x, y) -> factor $ normalPdf (slope * x + intercept) (sqrt noise) y)
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

docs/source/index.rst
```diff
@@ -0,0 +1,19 @@
+.. monad-bayes documentation master file, created by
+   sphinx-quickstart on Fri Dec 17 18:38:02 2021.
+   You can adapt this file completely to your liking, but it should at least
+   contain the root `toctree` directive.
+
+Welcome to monad-bayes's documentation!
```
idontgetoutmuch: I am no grammar expert but I think you could say monad-bayes' (since you already have an s)

docs/source/index.rst
```diff
@@ -0,0 +1,19 @@
...
+=======================================
+
+Monad-bayes is a library for doing probabilistic programming in Haskell, using probability monads. The emphasis is on specifying advanced inference algorithms in a compositional (and therefore succinct) way.
```
idontgetoutmuch: Do we need to say "using probability monads"? I think without explaining this, it doesn't really add very much. How about

Monad-bayes is a library for doing probabilistic programming in Haskell. Not only can you specify your model independent of the inference method as you can in all probabilistic programming languages but you can compose inference methods to create new ones. For example to use Markov Chain Monte Carlo (MCMC) to infer parameters for a model on which you are using Sequential Monte Carlo (SMC aka Particle Filtering) for inference assuming that you know the parameters in one line.

Hmmm that's still too much of a mouthful for a sentence that's supposed to grab your attention. My attention was definitely grabbed by the ability to compose in Adam's original paper.

BTW I now know that e.g. composing MCMC over SMC actually samples from the posterior not just an approximation to the posterior. I wonder if we should write that somewhere.

My assumption is that this documentation is for folks who have used e.g. Stan or PyMC and are curious how it is done in a functional way / in Haskell. If that's not the audience then we should probably have more background on Bayesian stats.

docs/source/probprog.md
idontgetoutmuch: We should say something about the model e.g.

$$
y_{n}=\alpha+\beta x_{n}+\epsilon_{n} \quad \text { where } \quad \epsilon_{n} \sim \operatorname{normal}(0, \sigma)
$$

idontgetoutmuch (changes requested):
https://github.com/tweag/monad-bayes/blob/8509452f28b9d55aa4165a6281e73d834a2923e8/docs/source/usage.md?plain=1#L432. I think this needs a bit more explanation e.g.

```
instance MonadSample m => MonadSample (Sequential m) where
  random = lift random
  bernoulli = lift . bernoulli
  categorical = lift . categorical
```

Also why are there only 3 methods?

PR #135, opened 2022-04-26
Port lazyppl monad into monad-bayes

reubenharry:
It was pretty straightforward to port the `Prob` monad into monad-bayes. I also ported over the lazy sampling algorithm and the weighted importance sampling algorithms (but not yet any of the mh variants). 

`Meas` doesn't even need to be defined explicitly in monad-bayes, because `Meas ~ Weighted Prob`. 

I *think* I ported the code in a semantics preserving way, but I'd want to check more carefully before merging.


idontgetoutmuch:
@reubenharry this seems to be lots of different changes all in one big PR. Can we split it into bite-sized PRs so that I can give appropriate attention to each change? Many thanks :-)

PR #140, opened 2022-05-09
Stratified sampling

idontgetoutmuch:
I think this is uncontroversial but can someone review it? There didn't seem to be many tests for systematic resampling but I copied what there was. I think the tests could be improved but I think we should make that a separate issue / PR.

### DIFF
--- a/src/Control/Monad/Bayes/Inference/SMC.hs
+++ b/src/Control/Monad/Bayes/Inference/SMC.hs
@@ -67,6 +67,8 @@ smcSystematic ::
   Population m a
 smcSystematic = sir resampleSystematic
 
+-- | Sequential Monte Carlo with stratified resampling at each timestep.
+-- Weights are not normalized.
 smcStratified ::
   MonadSample m =>
   -- | number of timesteps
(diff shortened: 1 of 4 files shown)


reubenharry (commented):
Agreed that it looks uncontroversial. My only request would be a comment describing what the stratified sampling method does on an intuitive level, and why its used (optionally with a link to a relevant resource). I can also add this to the docs.

idontgetoutmuch:
> Agreed that it looks uncontroversial. My only request would be a comment describing what the stratified sampling method does on an intuitive level, and why its used (optionally with a link to a relevant resource). I can also add this to the docs.

By that reasoning we should also have an explanation of how systematic resampling works - I can try to do both.

reubenharry:
> > Agreed that it looks uncontroversial. My only request would be a comment describing what the stratified sampling method does on an intuitive level, and why its used (optionally with a link to a relevant resource). I can also add this to the docs.
> 
> By that reasoning we should also have an explanation of how systematic resampling works - I can try to do both.

Ah, thanks! I wanted to do that, but didn't understand it well enough to.

reubenharry (approved):

PR #148, opened 2022-06-17
A draft for discussion

idontgetoutmuch:
@reubenharry I have put a draft here for us to think about. I haven't finished the transition as

1. Skolem variables escaping their scope in the `ST` monad (not a problem with only one RNG as unification sorts that out).
2. What do we do about defaults? I think for now we should use the RNG from `random` (*not* from `mwc-random`) as it is a lot faster. We can think about enhancing the interface later to provide more flexibility for the user (to choose their favourite RNG). I think this also sort out the Skolem problem.

Thoughts?

### DIFF
--- a/benchmark/SSM.hs
+++ b/benchmark/SSM.hs
@@ -9,21 +9,26 @@ import Control.Monad.Bayes.Sampler
 import Control.Monad.Bayes.Weighted
 import Control.Monad.IO.Class
 import NonlinearSSM
+import System.Random.Stateful
 
 main :: IO ()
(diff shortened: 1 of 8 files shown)


idontgetoutmuch:
@reubenharry all tests and benchmarks now run but we (I?) should set things up so we benchmark both the RNG in `mwc-random` and the RNG in `random`. I believe the latter to be faster.

Also this is a breaking change. We could try and make it non-breaking. But I don't know whether that is worth it. Maybe we could add migration advice to the Release Notes?

PR #149, opened 2022-06-17
Integrator, notebooks and helper functions

reubenharry:
A variety of updates, see changelog. Most prominent are:

- a new MonadSample instance called `Integrator` along with docs and tests for it
- notebooks which work. Using nix flakes from @idontgetoutmuch to run
- many more tests, e.g. that sampler gets right mean and variance
- use of foldl package to compute sample statistics
- a variety of models from the models folder fixed up and prettified. Some rewritten with pipes
- new Bayesian data type added to Control.Monad.Bayes.Class
- readme updated


reubenharry:
@idontgetoutmuch Do you know if the CI built failure seen above is a nix thing? It looks like a nix thing, and since I copied over your nix-flake (so I could work with the notebooks) I suspect that might have done something. Thoughts? 

Other than that, this is ready to review.

idontgetoutmuch (changes requested):
LGTM apart from CI no longer works.

reubenharry:
@idontgetoutmuch Any luck working out what the flake issue was here?

idontgetoutmuch (approved):

PR #155, opened 2022-07-05
Disable lint for now

idontgetoutmuch:
Until we can work out why `.hlint.yaml` is being ignored.


idontgetoutmuch:
@reubenharry ^

reubenharry (approved):

PR #156, opened 2022-07-05
Multivariate normal ii

idontgetoutmuch:
Should now pass CI

### DIFF
--- a/test/TestDistribution.hs
+++ b/test/TestDistribution.hs
@@ -1,10 +1,15 @@
 {-# LANGUAGE ImportQualifiedPost #-}
 {-# LANGUAGE Trustworthy #-}
 
-module TestDistribution (passed1) where
+module TestDistribution
+  ( passed1,
+    passed2,
+    passed3,
+  )
+where
 
 import Control.Monad (replicateM)
-import Control.Monad.Bayes.Class (mvNormal)
+import Control.Monad.Bayes.Class (MonadSample, mvNormal)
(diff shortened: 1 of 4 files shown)


reubenharry:
Looks good to me, pending tests, changelog and version bump. One test that comes to mind is looking at the marginals of e.g. a non-covariate multivariate normal.

test/TestDistribution.hs
```diff
@@ -0,0 +1,30 @@
...
+import Control.Monad.Bayes.Class (mvNormal)
+import Control.Monad.Bayes.Sampler
+import Control.Monad.Identity (runIdentity)
+import Control.Monad.State (evalStateT)
+import Data.Matrix (fromList)
+import Data.Vector qualified as V
+import System.Random.MWC (toSeed)
+
+passed1 :: Bool
+passed1 = runIdentity $ evalStateT r (toSeed (V.fromList [42, 1729]))
```
reubenharry: I'd advocate for using `sampleIOfixed` or `sampleSTfixed`, to make this cleaner.

reubenharry: Requesting a comment saying what this test is testing. And maybe a simple test that the estimated mean and covariance are the same as the true ones. (But I can also have a go at adding that test in a separate PR)

reubenharry (changes requested):

reubenharry (approved):

PR #157, opened 2022-07-05
Fix website link in README

fuzzypixelz:
Stumbled up this today and the website link seemed to be incorrect. I guessed it was `https://monad-bayes.netlify.app/` instead of `https://monad-bayes-site.netlify.app/` and that was true haha.

There doesn't seem to be be an `/about.html` page anymore though, so I dropped it from the link.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,11 @@
-# [Monad-Bayes](https://monad-bayes-site.netlify.app/about.html)
+# [Monad-Bayes](https://monad-bayes.netlify.app)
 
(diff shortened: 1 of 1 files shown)


idontgetoutmuch (approved):

reubenharry:
@fuzzypixelz Thanks for this! There's actually now also https://monad-bayes-site.netlify.app/_site/about.html, so I may make a PR to update it again, but thanks for catching this bad link!

PR #159, opened 2022-07-11
Prim monad

idontgetoutmuch:


benchmark/Speed.hs
```diff
@@ -131,10 +131,7 @@ samplesBenchmarks e lrData hmmData ldaData = benchmarks
 
 main :: IO ()
 main = do
-  g <- createSystemRandom
-  writeFile "speed-length.csv" ""
-  writeFile "speed-samples.csv" ""
-  writeFile "raw.dat" ""
+  g <- newIOGenM (mkStdGen 1729)
```
reubenharry, line 134: Can we make a synonym for `newIOGenM (mkStdGen 1729)`. It crops up a lot. Maybe `fixedGen`?

src/Control/Monad/Bayes/Sampler.hs
```diff
@@ -43,97 +44,127 @@ import Control.Monad.Bayes.Class
...
-    save,
-  )
+import Numeric.Log (Log (..))
 import System.Random.MWC.Distributions qualified as MWC
+import System.Random.Stateful (IOGenM, STGenM, StatefulGen, StdGen, mkStdGen, newIOGenM, newSTGenM, uniformRM)
 
 -- | An 'IO' based random sampler using the MWC-Random package.
-newtype SamplerIO a = SamplerIO (ReaderT GenIO IO a)
-  deriving newtype (Functor, Applicative, Monad, MonadIO)
+newtype SamplerIO g a = SamplerIO (StatefulGen g IO => ReaderT g IO a)
```
reubenharry: Why not just keep this compositional, as `Sampler g IO a` or (changing the variable order) `Sampler IO g a`?

reubenharry: That is, why not avoid the type synonym `SamplerIO` altogether?

src/Control/Monad/Bayes/Sampler.hs
```diff
@@ -43,97 +44,127 @@ import Control.Monad.Bayes.Class
...
+sampleIOwith :: StatefulGen r IO => SamplerIO r a -> r -> IO a
 sampleIOwith (SamplerIO m) = runReaderT m
 
-fromSamplerST :: SamplerST a -> SamplerIO a
-fromSamplerST (SamplerST m) = SamplerIO $ mapReaderT stToIO m
+sampleWith :: (StatefulGen r m) => Sampler r m a -> r -> m a
+sampleWith (Sampler m) = runReaderT m
+
+instance MonadSample (SamplerIO g) where
+  random = SamplerIO ((\s -> ask >>= lift . s) (uniformRM (0, 1))) -- FIXME: There is a special function for this
```
reubenharry: FIXME

src/Control/Monad/Bayes/Sampler.hs
```diff
@@ -43,97 +44,127 @@ import Control.Monad.Bayes.Class
...
+
+instance MonadSample (Sampler g m) where
+  random = Sampler (fromMWC' (uniformRM (0, 1))) -- FIXME: There is a special function for this
```
reubenharry: FIXME

src/Control/Monad/Bayes/Sampler.hs
```diff
@@ -43,97 +44,127 @@ import Control.Monad.Bayes.Class
...
 
-instance MonadSample SamplerIO where
-  random = fromSamplerST random
+  bernoulli p = Sampler (fromMWC' $ MWC.bernoulli p)
+  categorical ps = Sampler (fromMWC' $ MWC.categorical ps)
+  geometric p = Sampler (fromMWC' $ MWC.geometric0 p)
 
 -- | An 'ST' based random sampler using the @mwc-random@ package.
-newtype SamplerST a = SamplerST (forall s. ReaderT (GenST s) (ST s) a)
+newtype SamplerST g a = SamplerST (forall s. StatefulGen g (ST s) => ReaderT g (ST s) a)
```
reubenharry: Why not avoid the newtype, and just have `Sampler g ST a`?

src/Control/Monad/Bayes/Sampler.hs
```diff
@@ -43,97 +44,127 @@ import Control.Monad.Bayes.Class
...
-fromMWC :: (forall s. GenST s -> ST s a) -> SamplerST a
-fromMWC s = SamplerST $ ask >>= lift . s
+-- | Like 'sampleST' but with a custom pseudo-random number generator.
+sampleSTwith :: StatefulGen r (ST s) => SamplerST r a -> r -> ST s a
+sampleSTwith (SamplerST s) = runReaderT s
+
+fromMWC :: (g -> (ST s) a) -> ReaderT g (ST s) a
+fromMWC = (\s -> ask >>= lift . s)
+
+fromMWC' :: Monad m => (a -> m b) -> ReaderT a m b
```
reubenharry: Comments on `fromMWC` and `fromMWC'` would be nice

src/Control/Monad/Bayes/Sampler.hs
```diff
@@ -43,97 +44,127 @@ import Control.Monad.Bayes.Class
...
+fromMWC' = (\s -> ask >>= lift . s)
 
-instance MonadSample SamplerST where
-  random = fromMWC System.Random.MWC.uniform
+instance MonadSample (SamplerST g) where
+  random = SamplerST (fromMWC (uniformRM (0, 1))) -- FIXME: There is a special function for this
```
reubenharry: FIXME

test/TestInference.hs
```diff
@@ -25,32 +25,38 @@ import Control.Monad.Bayes.Inference.SMC
...
 
 sprinkler :: MonadInfer m => m Bool
 sprinkler = Sprinkler.soft
 
 -- | Count the number of particles produced by SMC
 checkParticles :: Int -> Int -> IO Int
 checkParticles observations particles =
-  sampleIOfixed (fmap length (runPopulation $ smcMultinomial observations particles Sprinkler.soft))
+  newIOGenM (mkStdGen 1729)
+    >>= sampleIOwith (fmap length (runPopulation $ smcMultinomial observations particles Sprinkler.soft))
```
reubenharry: Why not use `sampleIOfixed`?

reubenharry (commented):
Looks good to me in general. I left various low level comments. High level comments would be:

- we should definitely run the benchmarks, to make sure things aren't slower
- brief documentation in the changelog of what we changed
- explanation in the docs of how to use the new functions would be great (I'm also happy to have a go at that, but it would be nice to do it as part of this PR). Included in that would be examples of using different RNGs, and a comment on what to use for what. E.g., when do I use ST and when do I use IO?
- Since this is a breaking change, it might be nice to quickly follow it with the other breaking change I was considering, which was to do some renaming. These potential renamings would include:
- `sampleIO` -> `sampler`
- `runWeighted` -> `weighted`
- `runPopulation` -> `population`
- `sis` -> `sequentially`

benchmark/Speed.hs
idontgetoutmuch: Great feedback :-) thanks

Do you have a rationale for the name changes? I am particularly against sis -> sequentially as sis has a particular meaning in the world of SMC.

src/Control/Monad/Bayes/Sampler.hs
idontgetoutmuch: I think it was for backward compatibility but I am happy to remove it.

idontgetoutmuch:
> * explanation in the docs of how to use the new functions would be great (I'm also happy to have a go at that, but it would be nice to do it as part of this PR). Included in that would be examples of using different RNGs, and a comment on what to use for what.

There is some way of getting code in the documentation to run. Do you know how to do this?

test/TestInference.hs
idontgetoutmuch: Done

src/Control/Monad/Bayes/Sampler.hs
idontgetoutmuch: Done

src/Control/Monad/Bayes/Sampler.hs
idontgetoutmuch: Done

src/Control/Monad/Bayes/Sampler.hs
idontgetoutmuch: Done

src/Control/Monad/Bayes/Sampler.hs
idontgetoutmuch: Done

benchmark/Speed.hs
reubenharry: ah ok, I didn't know that. We can leave it

src/Control/Monad/Bayes/Sampler.hs
```diff
@@ -43,97 +40,52 @@ import Control.Monad.Bayes.Class
...
 
-fromSamplerST :: SamplerST a -> SamplerIO a
-fromSamplerST (SamplerST m) = SamplerIO $ mapReaderT stToIO m
+instance Applicative (Sampler g m) where
+  pure x = Sampler $ pure x
+  (Sampler f) <*> (Sampler x) = Sampler $ f <*> x
 
-instance MonadSample SamplerIO where
-  random = fromSamplerST random
+runSampler :: StatefulGen g m => Sampler g m a -> ReaderT g m a
```
reubenharry: Minor, but we could put this in the definition of `Sampler` as: `newtype Sampler g m a = Sampler { runSampler :: StatefulGen g m => ReaderT g m a}`

reubenharry:
@idontgetoutmuch One thing it would be nice to retain is `sampleIO` (which I might subsequently rename to `sampler`, which uses IO to create a random seed. I pushed an update with the following, lmk if it looks right to you:

```haskell
sampleIO :: Sampler (IOGenM StdGen) IO a -> IO a
sampleIO x = initStdGen >>= newIOGenM >>= sampleWith x
```

src/Control/Monad/Bayes/Sampler.hs
reubenharry: nvm, it didn't work cause of the constraint, as you probably already knew

reubenharry:
@idontgetoutmuch I'm seeing a pretty large slowdown in the benchmarks for the PrimMonad branch. I think we should definitely debug that before proceeding with this branch.

idontgetoutmuch:
> @idontgetoutmuch I'm seeing a pretty large slowdown in the benchmarks for the PrimMonad branch. I think we should definitely debug that before proceeding with this branch.

Huh - well that is very irksome - it really should speed things up

reubenharry:
Yeah, I imagine there's just a bug, or perhaps something needs to be specialized. I don't see a huge difference between the raw samplers (just timing the sampling of 100000 samples), so it's perhaps a function of changes in the benchmarking code, or elsewhere.

reubenharry:
If I do 

```haskell

import Text.Printf
import System.CPUTime

time :: IO t -> IO t
time a = do
    start <- getCPUTime
    v <- a
    end   <- getCPUTime
    let diff = (fromIntegral (end - start)) / (10^12)
    printf "Computation time: %0.3f sec\n" (diff :: Double)
    return v

time $ sampleIOfixed $ void $ replicateM 100000 random
```

I get 

> Computation time: 0.134 sec

which is comparable to the main branch.

idontgetoutmuch:
> Yeah, I imagine there's just a bug, or perhaps something needs to be specialized. I don't see a huge difference between the raw samplers (just timing the sampling of 100000 samples), so it's perhaps a function of changes in the benchmarking code, or elsewhere.

I think it could be the way I set up benchmarking. I am just about to benchmark the linear regression example you came up with just using +RTS -s.

idontgetoutmuch:
```
git branch
* master

  INIT    time    0.000s  (  0.002s elapsed)
  MUT     time    0.975s  (  0.985s elapsed)
  GC      time    0.173s  (  0.187s elapsed)
  EXIT    time    0.000s  (  0.004s elapsed)
  Total   time    1.148s  (  1.179s elapsed)

git branch
* PrimMonad

  INIT    time    0.000s  (  0.003s elapsed)
  MUT     time    0.975s  (  0.988s elapsed)
  GC      time    0.173s  (  0.190s elapsed)
  EXIT    time    0.000s  (  0.012s elapsed)
  Total   time    1.148s  (  1.193s elapsed)
```

idontgetoutmuch:
So I think it must be something to do with the `ST` monad which I think is used in the performance tests. I won't be able to look further until at least Friday 22 July.

```
{-# LANGUAGE BlockArguments      #-}
{-# OPTIONS_GHC -Wall            #-}

module Main(main) where

import Control.Monad.Reader

import Control.Monad.Bayes.Traced
import Control.Monad.Bayes.Weighted

paramPriorRegression :: MonadSample m => m (Double, Double, Double)
paramPriorRegression = do
    slope <- normal 0 2
    intercept <- normal 0 2
    noise <- gamma 1 1 -- 4 4
    return (slope, intercept, noise)

regression :: (MonadInfer m) => [Double] -> [Double] -> m (Double, Double, Double)
regression xs ys = do
    (slope, intercept, noise) <- paramPriorRegression
    _ <- forM (zip xs ys) \(x, y) -> factor $ normalPdf (slope * x + intercept) (sqrt noise) y
    return (slope, intercept, noise)

syntheticData :: MonadSample m => Int -> m [(Double, Double)]
syntheticData n = do
  let xs = map fromIntegral [1 .. n]
      ys = map (3*) xs
  eps <- replicateM n $ normal 0 0.1
  let zs = zipWith (+) ys eps
  return $ zip xs zs

main :: IO ()
main = do
  xys <- sampleIO (syntheticData 10)
  let (xs, ys) = unzip xys
  print xs
  print ys
  putStrLn "\n"

  mhRunsRegression <- sampleIO $ prior $ mh 100000 $ regression xs ys
  let ss = map (\(x, _, _) -> x) mhRunsRegression
  let is = map (\(_, y, _) -> y) mhRunsRegression
  let ns = map (\(_, _, z) -> z) mhRunsRegression
  print $ (/ 50000.0) $ Prelude.sum $ drop 50000 ss
  print $ (/ 50000.0) $ Prelude.sum $ drop 50000 is
  print $ (/ 50000.0) $ Prelude.sum $ drop 50000 ns
```

```
executable perftestmain
  main-is:            PerfTestMain.hs
  hs-source-dirs:     .
  other-modules:
  build-depends:
      base,
      monad-bayes,
      mtl,
      mwc-random,
      random
```

reubenharry (approved):

PR #168, opened 2022-08-14
Stormer Verlet

idontgetoutmuch:
I've put a test / example in `test` but would it be better to put it in the haddocks?


reubenharry (approved):
Looks great

PR #171, opened 2022-08-16
Generalize from double

reubenharry:
PR to generalize MonadSample and MonadInfer from Double. Also has beginnings of implementation of HMC, in Control.Monad.Bayes.Traced.Grad. Quite a few things to fix.


benchmark/Single.hs
```diff
@@ -2,9 +2,11 @@
 {-# LANGUAGE ImportQualifiedPost #-}
```
idontgetoutmuch, line 2: Why is everything commented out?

benchmark/SSM.hs
```diff
@@ -1,35 +1,43 @@
 module Main where
```
idontgetoutmuch, line 1: I thought we had already made the changes to use a record to contain the various options?

benchmark/Speed.hs
```diff
@@ -41,7 +39,7 @@ instance Show Model where
   show (HMM xs) = "HMM" ++ show (length xs)
   show (LDA xs) = "LDA" ++ show (length $ head xs)
 
-buildModel :: MonadInfer m => Model -> m String
+buildModel :: MonadInfer n m => Model -> m String
```
idontgetoutmuch, line 42: This extra parameter seems pretty major to me. Can we explain it in the PR itself? I think we should be a bit more disciplined in our PRs going forward (I include myself here and I may be the biggest culprit).

benchmark/Speed.hs
```diff
@@ -83,8 +88,8 @@ systems =
   [ MonadBayes
   ]
 
-lengthBenchmarks :: Env -> [(Double, Bool)] -> [Double] -> [[T.Text]] -> [Benchmark]
-lengthBenchmarks e lrData hmmData ldaData = benchmarks
+lengthBenchmarks :: [(Double, Bool)] -> [Double] -> [[T.Text]] -> [Benchmark]
```
idontgetoutmuch, line 91: Not that I loved `Env` but why is it being removed?

idontgetoutmuch (changes requested):
Are these large changes e.g. https://github.com/tweag/monad-bayes/pull/171/files#diff-0c0b7688a56339683b2983c70870dfcdc7e9746176d51e7fc6a906fed38528ce because we are still storing the generated html in the repo or is this just the source text for the html and this is a one off?

models/ConjugatePriors.hs
```diff
@@ -34,33 +34,33 @@ gammaNormalAnalytic (a, b) points = gamma a' (recip b')
     b' = b + sum (fmap (** 2) points) / 2
 
 -- | Posterior on beta after the bernoulli sample
-betaBernoulliAnalytic :: (MonadInfer m, Foldable t) => BetaParams -> t Bool -> m Double
-betaBernoulliAnalytic (a, b) points = beta a' b'
-  where
-    (n, s) = fold (liftA2 (,) F.length (F.premap (\case True -> 1; False -> 0) F.sum)) points
-    a' = a + s
-    b' = b + fromIntegral n - s
+-- betaBernoulliAnalytic :: (MonadInfer n m, Foldable t) => BetaParams -> t Bool -> m n Double
```
idontgetoutmuch, line 37: Why is everything commented out? Is this PR not really ready for review?

idontgetoutmuch (changes requested):

PR #174, opened 2022-08-26
Piped mcmc

reubenharry:
This is MCMC with pipes, but without the terminal user interface (TUI).


src/Control/Monad/Bayes/Inference/MCMC.hs
```diff
@@ -11,12 +11,17 @@
 -- Portability : GHC
 module Control.Monad.Bayes.Inference.MCMC where
 
-import Control.Monad.Bayes.Class (MonadSample)
+import Control.Monad.Bayes.Class
+import Control.Monad.Bayes.Sampler.Strict
 import qualified Control.Monad.Bayes.Traced.Basic as Basic
-import Control.Monad.Bayes.Traced.Common (burnIn)
+import Control.Monad.Bayes.Traced.Common
```
idontgetoutmuch: I don't want to cause a religious war but I think for production code it's good to have explicit imports as then it's easier for a maintainer to see where a function comes from. Of course, during development this is a pain. Thoughts?

src/Control/Monad/Bayes/Inference/MCMC.hs
```diff
@@ -11,12 +11,17 @@
...
 import qualified Control.Monad.Bayes.Traced.Dynamic as Dynamic
 import qualified Control.Monad.Bayes.Traced.Static as Static
 import Control.Monad.Bayes.Weighted
+import Control.Monad.Trans (lift)
```
idontgetoutmuch, line 21: And here you have made the import explicit.

idontgetoutmuch (changes requested):
I don't see any tests? Maybe not necessary for this PR but is there a performance difference?

src/Control/Monad/Bayes/Inference/MCMC.hs
reubenharry: I agree

reubenharry: Re tests, I'm not using the piped version as the default one, precisely cause I haven't yet benchmarked.

idontgetoutmuch (approved):

PR #177, opened 2022-08-29
Generalize from double: with type family

reubenharry:


src/Control/Monad/Bayes/Class.hs
```diff
@@ -88,76 +91,89 @@ import Statistics.Distribution.Geometric (geometric0)
...
   random ::
     -- | \(\sim \mathcal{U}(0, 1)\)
-    m Double
+    m (Real m)
 
   -- | Draw from a uniform distribution.
   uniform ::
     -- | lower bound a
-    Double ->
+    (Real m) ->
```
turion, line 114:
```suggestion
    Real m ->
```

reubenharry:
@mknorps @idontgetoutmuch If we still want HMC, we'll need to merge/rebase this branch: it generalizes from Double in a way that allows for automatic differentiation.

turion:
@reubenharry What is missing? Do you need any help? We're working around `statistics` only supporting `Double` right now, right?

idontgetoutmuch:
Do we need `statistics`? Would `mwc-random` be enough? I am sure @Shimuuar would accept a PR to generalise the types e.g. here https://hackage.haskell.org/package/mwc-random-0.15.0.2/docs/System-Random-MWC-Distributions.html provided we didn't affect performance.

BTW we should also check performance before and after generalising `Double` to `Num a` (or whatever).

turion:
`mwc-random` also pins the implementations of the distributions to `Double`.

idontgetoutmuch:
> `mwc-random` also pins the implementations of the distributions to `Double`.

I understand but I think the maintainer would be sympathetic to generalising it provided we did not impact performance.

reubenharry:
@turion So the `ad` library supports the inverse error function, so we can at least currently do normal distributions without `statistics`. Less so for the other distributions in the `MonadSample` class. I don't know if we mind the sampler working in terms of `Double`, but if we do, we'd indeed want to request a change to `mwc-random` or move to something else.

Shimuuar:
I'm not against generalization but it's kind of tricky. Some generators would be easy: exponential do not require anything beyond `log`. Others may be trickier. Generator for normal distribution uses ziggurat method and corresponding tables are only available for Doubles, some may use `math-functions` which provide only versions for Doubles as well

turion:
> Generator for normal distribution uses ziggurat method and corresponding tables are only available for Doubles

Why is that? It seems to me that the table is calculated in Haskell as well, so one could calculate several tables, possibly creating a type class for this.

Shimuuar:
Why ziggurat? It's fast. Last time I run benchmarks (several years before random 1.2) it was only tens of percent slower than generation of Double. I didn't compare it with inversion method though.

Creating type class is possible but that's road leads to type class per distribution.

Issue #165, opened 2022-08-09
Create a nix flake that supports both the package and the associated notebooks


idontgetoutmuch:
Currently we have two nix flakes: one to build and test the library and one for running the Jupiter notebooks. Switching between these e.g. by changing branches or updating the flakes is a pain.

We'd like just one flake that provides a dev environment, allows us to to all the checks locally that get run via CI and also be able to run the notebook examples.

The alternative is stay with the manual solution.

aleeusgr:
I will work on this too, you're welcome to guide me.

idontgetoutmuch:
One of my colleagues advised:

> Make a new flake which has your two flakes as inputs, then compose as needed

> E.g. if you have a flake which exports `packages.x86_64-linux.hello` and another flake which exports `packages.x86_64-linux.goodbye`, you can do something like

```
{
  inputs.nixpkgs = "github:nixos/nixpkgs";
  inputs.a-flake = "github:you/a-flake";
  inputs.another-flake = "github:you/another-flake";
  outputs = { self, a-flake, another-flake }: {
    packages.x86_64-linux.a-conversation =
      nixpkgs.legacyPackages.x86_64-linux.runCommand "conversation" { }
      "${a-flake.packages.x86_64-linux.hello} > $out; ${another-flake.packages.x86_64-linux.goodbye} >> $out";
  };
}
```

idontgetoutmuch:
The two flakes are

https://github.com/tweag/monad-bayes/blob/notebooks/flake.nix

and

https://github.com/tweag/monad-bayes/blob/master/flake.nix

With the former you can do

```
nix develop --accept-flake-config --system x86_64-darwin
jupyter-lab
```

and with the latter you can do

```
nix develop --accept-flake-config --system x86_64-darwin
```

and

```
nix --print-build-logs build .#pre-commit
```

idontgetoutmuch:
@aleeusgr good luck - neither I nor @reubenharry have very high nix fu but we will help as best we can.

aleeusgr:
@idontgetoutmuch thank you, awesome.
me neither, but I'd love to improve. 

You want to have `flake.nix`  in root such that `nix develop` allows running `jupyter-lab` ?

aleeusgr:
```
cd notebooks
nix develop --accept-flake-config --system x86_64-linux
jupyter-lab
```


`jupyter-lab: command not found`

idontgetoutmuch:
Which branch are you on?

[...]

PR #187, opened 2022-09-07
to test run `jupyter-lab` inside nix shell in master flake.

aleeusgr:
Hey, I'm working on https://github.com/tweag/monad-bayes/issues/165
tested on my machine. 

plz notice that the url in the inputs is to my fork. If I change it to tweag, I get an error:
> the lockfile is broken, it contains merge conflict markers https://github.com/tweag/monad-bayes/blob/notebooks/flake.lock#L242

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -567,11 +567,11 @@
     },
     "nixpkgs_2": {
       "locked": {
-        "lastModified": 1661720780,
-        "narHash": "sha256-AJNGyaB2eKZAYaPNjBZOzap87yL+F9ZLaFzzMkvega0=",
+        "lastModified": 1662019588,
+        "narHash": "sha256-oPEjHKGGVbBXqwwL+UjsveJzghWiWV0n9ogo1X6l4cw=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "a63021a330d8d33d862a8e29924b42d73037dd37",
+        "rev": "2da64a81275b68fdad38af669afeda43d401e94b",
         "type": "github"
       },
       "original": {


flake.nix
```diff
@@ -19,6 +19,7 @@
     pre-commit-hooks.inputs.nixpkgs.follows = "nixpkgs";
     pre-commit-hooks.inputs.flake-utils.follows = "flake-utils";
     haskell-nix-utils.url = "github:TerrorJack/haskell-nix-utils";
+    jupyter-flake.url = "git+https://github.com/aleeusgr/monad-bayes?ref=notebooks";
```
idontgetoutmuch: Why is there a reference to your fork?

idontgetoutmuch (changes requested):
This is great :-) thanks very much - it works really well but I saw a reference to your clone?

EDIT: I don't know how to resolve the merge conflicts.

flake.nix
idontgetoutmuch: Huh - I just saw your comment above - let me try and ask the experts at tweag

idontgetoutmuch: I just changed the url to tweag and it works for me (that's not how nix is supposed to work)

idontgetoutmuch:
@aleeusgr I think you need to rebase on to the latest master? I tried doing a merge on my computer but I get lots of changes that are nothing to do with your PR.

flake.nix
aleeusgr: I think it has to do with cached versions of stuff in nix store. after rebase the error seems to be gone.

Issue #189, opened 2022-09-13
T in the transformer names


reubenharry:
Sequential -> SequentialT
Traced -> TracedT
Population -> PopulationT
run[X] -> run[X]T

This helps Haskellers understand what is going on better.
```[tasklist]
### Tasks
- [ ] Sequential
- [ ] Traced
- [ ] Population
- [ ] Weighted
- [ ] Density
- [ ] Sampler
- [ ] run* functions
```

aleeusgr:
I'd love to help, but this does not make sense to me.

If you'd shared a doc on what is "transformer", how to find them in the code, I'd think of this as a task I could do.

Thanks.

reubenharry:
Hi. Sure, I can clarify! 

How familiar are you with Haskell? If not, it's actually ok for this one, because basically what I'm suggesting is just a renaming, as follows:

Everywhere "Sequential" appears in the code, we replace it with "SequentialT". So the type `Sequential` becomes `SequentialT`, the constructor `Sequential` also becomes `SequentialT` and `runSequential` becomes `runSequentialT`. 

Similarly for `Population` and `Traced`. 

I think search and replace would mostly work for this.

aleeusgr:
I'm ok with Haskell, not a brightest of minds but I'm learning:)
there's only four of them?
Ok, search and replace, on it.

`grep -rnw -e 'Sequential' --include \*.hs`

aleeusgr:
`find . -type f -exec sed -i 's/Sequential/SequentialT/g' {} \;`

https://stackoverflow.com/questions/15402770/how-to-search-and-replace-using-grep

Hm, It breaks git repo...

`find ./*.hs -type f -exec sed -i 's/Sequential/SequentialT/g' {} \;` ??

aleeusgr:
`find . -name "*.hs" -type f -exec sed -i 's/<oldname>/<newname>/g' {} \;`

also not to forget to rename .hs files and tests; update monad-bayes.cabal

aleeusgr:
Hey @reubenharry 
I figured out Sequential, Traced and Population - `stack test` is passing.
but what's up with run[X]?

`$ grep -rnw -e 'run' --include \*.hs`
returns 
```
test/TestSequentialT.hs:22:finishedTwoSync n = finished (run n twoSync)
test/TestSequentialT.hs:24:    run 0 d = d
test/TestSequentialT.hs:25:    run k d = run (k - 1) (advance d)
[...]

PR #196, opened 2022-09-18
Renaming transformers.

aleeusgr:
This is code for https://github.com/tweag/monad-bayes/issues/189

after, renaming files and .cabal manipulation is required.

Cheers.


src/Control/Monad/Bayes/PopulationT.hs
```diff
@@ -65,44 +65,44 @@ import Numeric.Log qualified as Log
...
-  lift = Population . lift . lift
+instance MonadTrans PopulationT where
+  lift = PopulationT . lift . lift
 
 -- | Explicit representation of the weighted sample with weights in the log
 -- domain.
-population, runPopulation :: Population m a -> m [(a, Log Double)]
-population (Population m) = runListT $ weighted m
+population, runPopulationT :: PopulationT m a -> m [(a, Log Double)]
+population (PopulationT m) = runListT $ weighted m
```
turion, line 77: It is more customary to have a name like `population` to be a constructor for the type. See e.g. https://hackage.haskell.org/package/transformers-0.6.0.4/docs/Control-Monad-Trans-Reader.html#v:reader. From that perspective I'd expect this type signature:

```haskell
population :: [a, Log Double] -> PopulationT m a
```

reubenharry: That's true. I think this can be deferred to a separate PR, but I agree. We'd then want `runSamplerIO`, `runWeighted` and so on.

reubenharry:
Thanks for doing this! This looks great to me, but I think it's worth waiting for the big notebooks PR first, since they are going to conflict, and I think this one will be easier to merge.

idontgetoutmuch:
@aleeusgr I am sorry this has sat here for so long - I am trying to determine what else needs to be done

aleeusgr:
I don't mind.

I'm sorry I didn't find time to help you in some other way: we came up with a project to improve our infrastructure, Cardano and Catalyst, and got funding to build an MVP, so I'm deep in my project!

I feel immediate benefits from doing this, and I see how this project works in the long run, and this makes it hard to put anything else in my scope at the moment.

turion (changes requested):
Hi :) thanks a lot for your work so far! Do you want to rebase it and pick it up? I'll guide you through the further PR process.

For starters: It seems that you have renamed _all_ occurrences of e.g. `Population` to `PopulationT`, but we really only want the _type_ to be renamed. For example, the module should still be called `Population`. Same goes for `Sequential`, and so on.

src/Control/Monad/Bayes/Inference/RMSMC.hs
```diff
@@ -3,14 +3,14 @@
...
+-- Description : Resample-Move SequentialT Monte Carlo (RM-SMC)
 -- Copyright   : (c) Adam Scibior, 2015-2020
 -- License     : MIT
 -- Maintainer  : leonhard.markert@tweag.io
 -- Stability   : experimental
 -- Portability : GHC
 --
--- Resample-move Sequential Monte Carlo (RM-SMC) sampling.
+-- Resample-move SequentialT Monte Carlo (RM-SMC) sampling.
 --
```
turion, line 14: These changes should be removed

aleeusgr: could you describe the proposed impact of this work?
as a research fellow without a project I have infinitely many hours to invest, but I value resourcefulness, efficiency and sustainability. 
Figuring out one line of code that would implement significant change and satisfy the requirements was fun and interesting task.
Cleaning up the mess is simpler with "close request" button.

Do you guys need a technical writer at Tweag? Because during an interview 6 months ago I came across a natural language specification that looked like the author hated every second of the process.

turion: > could you describe the proposed impact of this work?

You mean this PR here? It's about improving maintainability of the library and decreasing surprise for Haskell developer discovering this library, to enable them to program with less friction.

> Do you guys need a technical writer at Tweag?

About that I have no idea, but I guess Github is the wrong place to inquire that. I'd rather recommend heading to the careers page on the Tweag website.

Issue #197, opened 2022-09-23
Renaming `MonadSample` and `MonadInfer`


reubenharry:
I'm not a huge fan of the names `MonadSample` and `MonadInfer`. On the one hand, it's nice to keep aligned with the original paper. But on the other, they're a little misleading. `MonadSample` isn't inherently about sampling.

I'd proposed the names `MonadDistribution` and `MonadMeasure` respectively.

The downsides of these are:

- slightly longer

Finally I'd replace `MonadCond` with `MonadFactor`.

mknorps:
I the library there is already a more informative type signatures for some of the use cases (see [HERE](https://github.com/tweag/monad-bayes/blob/436eb3f7864d1026e23245dec7c94aebb439e5a5/src/Control/Monad/Bayes/Class.hs#L258)):
```
type Distribution a = forall m. MonadSample m => m a

type Measure a = forall m. MonadInfer m => m a

type Kernel a b = forall m. MonadInfer m => a -> m b
```
Do you find adding `Monad` prefix more Haskelly? Another option to go with the Haskell style is adding a suffix `M` indicating we are in a monadic context.
Then we would have `DistributionM`, `MeasureM`, etc.

reubenharry:
That's true - I added `Distribution` etc in Class.hs, but they're a bit restrictive because you can't write e.g. `a -> Distribution b` for a conditional distribution, and instead have to use `Kernel`. 

I think for typeclasses, the convention is `MonadX`, as in `MonadReader`, `MonadState`, etc, but for functions, it's `xM`.

mknorps:
You are right @reubenharry! I found a summary on [Kowainik blog](https://kowainik.github.io/posts/naming-conventions). As for adding `Monad` prefix for typeclasses it is used when a typeclass has mtl style effects, so all is good.

PR #179, opened 2022-08-30
Website

reubenharry:
Big PR, bumping the version to 1.0.0. Hosts monad-bayes website and docs, updates the notebooks, fixes smaller issues.

PR #198, opened 2022-09-30
Rename

reubenharry:
Re issue #197. Obviously this is quite a large breaking change, so would merit a new release. But I think especially for didaction purposes, the new names are much more clear. It's confusing to have to explain that `MonadSample` doesn't have anything inherently to do with sampling, or that `MonadInfer` doesn't have anything inherently to do with inference.


idontgetoutmuch:
How does this relate to https://github.com/tweag/monad-bayes/pull/196? We said there that we would commit https://github.com/tweag/monad-bayes/pull/179 first to avoid a more complicated rebase. So I assume we won't merge this until afterwards?

reubenharry:
> How does this relate to #196? We said there that we would commit #179 first to avoid a more complicated rebase. So I assume we won't merge this until afterwards?

That sounds right to me

idontgetoutmuch:
So I think this is now ok to merge? @reubenharry @mknorps?

reubenharry:
Needs rebase. I will do that now. Easier actually to just redo from main.

PR #206, opened 2022-10-07
Notebooks: sampling.ipynb

mknorps:
Small adjustments to Sampling notebook:
- add a histogram of samples to give more examples of split between the sampling part vs the distribution construction part and where it is applied.
- described usage of histogram on simple continuous distribution (map of constant weights)


mknorps:
@reubenharry Could you explain the weighted vs unweighted sampling a bit more? I do not understand why we need `weighted` in caseof a normal distribution:
image

reubenharry:
> @reubenharry Could you explain the weighted vs unweighted sampling a bit more? I do not understand why we need `weighted` in caseof a normal distribution: image

I think you're right, and `unweighted` would be fine. I believe they're equivalent here (since there were no factor statements), but `unweighted` is clearer, I agree.

mknorps:
> > @reubenharry Could you explain the weighted vs unweighted sampling a bit more? I do not understand why we need `weighted` in caseof a normal distribution: image
> 
> I think you're right, and `unweighted` would be fine. I believe they're equivalent here (since there were no factor statements), but `unweighted` is clearer, I agree.

The problem I have with this statement is that it needs  `weighted` or a `unweighted . weighted` functions at all. It is not clear for me:
- why we have to treat sampling from continuous distribution with additional weights by default and not having weights as an option?
- if the above is necessary due to implementation logic, could you explain why it is so?

reubenharry:
> > > @reubenharry Could you explain the weighted vs unweighted sampling a bit more? I do not understand why we need `weighted` in caseof a normal distribution: image
> > 
> > 
> > I think you're right, and `unweighted` would be fine. I believe they're equivalent here (since there were no factor statements), but `unweighted` is clearer, I agree.
> 
> The problem I have with this statement is that it needs `weighted` or a `unweighted . weighted` functions at all. It is not clear for me:
> 
> * why we have to treat sampling from continuous distribution with additional weights by default and not having weights as an option?
> * if the above is necessary due to implementation logic, could you explain why it is so?

I don't think we need `weighted`, or even `unweighted`. The only thing is that we need some weights to give to the histogram, so we could just supply those (something like `fmap (,1) (replicateM n model2)` (That ends up being what `weighted` does anyway, which is why the code is the way it is, but I agree it's confusing).

reubenharry (approved):
These changes look good to me!

PR #211, opened 2022-10-12
Use custom stream of trees type in lazy sampler

### LINKED ISSUE #184: Why use Lists for streams in the lazy sampler?
Just a small nitpick. Wouldn't it be clearer to use a stream type in the definition of trees?

```
data Stream a = Stream a (Stream a)
data Tree = Tree Double (Stream Tree)
```

Otherwise, it is too easy to mistake the `Tree` for a [rose tree](https://en.wikipedia.org/wiki/Rose_tree).

If you're interested, I can send a PR for this.

turion:
Fixes #184

### DIFF
--- a/src/Control/Monad/Bayes/Inference/Lazy/MH.hs
+++ b/src/Control/Monad/Bayes/Inference/Lazy/MH.hs
@@ -13,6 +13,7 @@ module Control.Monad.Bayes.Inference.Lazy.MH where
 import Control.Monad.Bayes.Sampler.Lazy
   ( Sampler (runSampler),
     Tree (..),
+    Trees (..),
     randomTree,
   )
 import Control.Monad.Bayes.Weighted (Weighted, weighted)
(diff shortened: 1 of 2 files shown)


reubenharry:
Why not use a library definition of a stream? Unnecessary overhead?

reubenharry (approved):

turion:
> Why not use a library definition of a stream? Unnecessary overhead?

Yes, I looked into several libraries. Some were well-built, but complex like https://hackage.haskell.org/package/streaming-0.2.3.1/docs/Streaming.html, some were simple enough like https://hackage.haskell.org/package/streams but seem not active enough to be a reliable dependency.

turion:
If you want, I can rewrite with `streaming`. I think that would be my first choice.

turion:
...but I think that at this point, a dependency would be unnecessary overhead. We're not using any fancy stream manipulation functions (yet).

PR #219, opened 2022-10-18
Rename

reubenharry:
MonadSample -> MonadDistribution
MonadCond -> MonadFactor
MonadInfer -> MonadMeasure

Possibly still not optimal, but hopefully much better than the very confusing names previously.


src/Control/Monad/Bayes/Class.hs
```diff
@@ -106,7 +106,7 @@ import Statistics.Distribution.Poisson qualified as Poisson
 import Statistics.Distribution.Uniform (uniformDistr)
 
 -- | Monads that can draw random variables.
-class Monad m => MonadSample m where
+class Monad m => MonadDistribution m where
```
mknorps, line 109: This comment is more about the definition of the `MonadDistribution` than the name itself, which I think is OK.

`MonadDistribution` contains few standard distributions and `random` function. Other functions like `draw`, `fromPMF`, `discrete` are tangled in particular distributions sampling but are outside of the typeclass. Is there a particular reason to design it that way?
What if we gathered all tools needed to draw samples from distributions in `MonadDistribution` and move concrete distributions like `nomal` outside of the class?

mknorps (approved):
The names are better indeed.
I have one question regarding design of `MonadDistribution`, but it does not affect the name changes directly.

src/Control/Monad/Bayes/Class.hs
mknorps: It may also be that I have metal model of encapsulation in OOP :)

reubenharry: This is indeed a very reasonable question, but there is a rationale to the design. Minimally, all that `MonadDistribution` needs is `random`, and then (most) other distributions can then be defined using `fmap f random` for some `f`. However, for reasons of efficiency, it's not always optimal to define e.g. `normal a b` as `fmap (inverseNormalCDF a b) random`, which is the default. We want to be able to override this for special cases, like the instance for `Sampler`, and so, we need to provide `normal` in the class itself. Or at least, I couldn't see a better solution.

mknorps: And then to maintain coherence, all other distributions are placed inside `MonadDistribution` as well?

reubenharry: Well, similar reasoning applies. If you look at the instance for `Sampler`, you'll see it gives its own implementations for most of the distributions provided by `MonadDistribution`, for reasons of efficiency. Those are the ones that need to be in the class (since otherwise, `Sampler` has no way of special-casing them).

PR #220, opened 2022-10-19
Notebooks: Bayesian.ipynb

mknorps:
Detailed review of Bayesian.ipynb tutorial.


mknorps:
@reubenharry
What do you mean by "an't always get the density of a sample"? 
Could you elaborate on the difference between `generative` and `likelihood`?
> Monad-Bayes exports a `Bayesian` datatype to make using Bayesian models easier. The idea is that you specify your model as a prior, generative model and a likelihood (separate from the generative model because you can't always get the density of a sample from a distribution).

reubenharry:
> @reubenharry What do you mean by "an't always get the density of a sample"? Could you elaborate on the difference between `generative` and `likelihood`?
> 
> > Monad-Bayes exports a `Bayesian` datatype to make using Bayesian models easier. The idea is that you specify your model as a prior, generative model and a likelihood (separate from the generative model because you can't always get the density of a sample from a distribution).

Yes, I didn't explain that well. What I mean is that from code like `normal 0 1`, there's no automatic way to transform it into `normalPdf 0 1`, or more generally, if I have `x :: MonadDistribution m => m a`, I can't automatically get `y :: a -> Log Double`.

reubenharry (approved):

Issue #127, opened 2022-03-15
Updated docs 


reubenharry:
- [x] documentation (of which a draft is https://monad-bayes.netlify.app/)
- [x] a tutorial, in notebook form, covering independent sampling, MCMC, particle filters, rmsmc, pmmh, smc2 (https://probmods.org/ is an excellent example of how to do this well (in that case the focus is on using probabilistic programming to model cognition) )
- [x] update the haddock docs with more comments (e.g. difference between different resamplers)
- [x] update hackage to link to the docs and tutorials

The broad goal is to make monad-bayes accessible to anyone with a beginner understanding of Haskell and a beginner understanding of Bayesian inference. That will require clear tutorials with simple working examples.

idontgetoutmuch:
This is great - I am writing up an example of PMMH independent of monad-bayes - I've put notes in pdf form here:

[odeModelExpanded.pdf](https://github.com/tweag/monad-bayes/files/8253569/odeModelExpanded.pdf)

The code is here:

1. https://github.com/idontgetoutmuch/abm-particle-filter/blob/mathematical-exposition/src/Data/PMMH.hs#L42
2. https://github.com/idontgetoutmuch/abm-particle-filter/blob/mathematical-exposition/src/Data/PMMH.hs#L93

I've also used [Stan](https://mc-stan.org/) to infer the parameters albeit for a slightly(?) different model (one that is fully deterministic once you know the parameters). This is included in the pdf above and in the git repo.

MMesch:
This looks great. Looking forward to the PR. I can iterate through it.

reubenharry:
Thanks for those notes! One of the more time-intensive parts of the docs that I haven't yet done is to have a worked through example of PMMH, so I may use your notes as reference material for writing that.

idontgetoutmuch:
@reubenharry I don't know how much mathematical detail you want to put in the docs. I think the Stan documentation is great but even that does not try to explain Hamiltonian Monte Carlo. Ah but it does here: https://mc-stan.org/docs/2_29/reference-manual/hamiltonian-monte-carlo.html although not particularly rigorously or in great detail. Anyway I am nowhere near finished explaining HMC let alone how one can use it to approximate the posterior (log) likelihood for use in MCMC (or even HMC). You can see what you think when I have finished. I am attempting to make it accessible to a widish audience while still remaining mathematically unsullied :-)

reubenharry:
[...]

PR #222, opened 2022-10-20
Make README visible on hackage

idontgetoutmuch:
As requested: https://github.com/tweag/monad-bayes/issues/127

### DIFF
--- a/monad-bayes.cabal
+++ b/monad-bayes.cabal
@@ -1,6 +1,6 @@
 cabal-version:      2.0
 name:               monad-bayes
-version:            1.0.0
+version:            1.1.0
 license:            MIT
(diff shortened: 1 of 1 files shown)


monad-bayes.cabal
```diff
@@ -131,7 +136,7 @@ executable example
 
   if flag(dev)
     ghc-options:
-      -O2 -Wall -Wcompat -Wincomplete-record-updates
```
reubenharry, line 134: Why remove O2?

idontgetoutmuch: The package uploader suggested it makes no difference and slows down compilation times.

reubenharry (approved):

PR #225, opened 2022-10-26
Notebooks: SMC.ipynb

mknorps:
This PR regards updating SMC.ipynb notebook with explanation the idea behind Sequential Monte Carlo based on present random walk example.

I aim at shifting the weight of the notebook from the random walk example to SMC and its implementation in monad-bayes.


mknorps:
@reubenharry could you help me explain the last plot (copy below)
image

I thought it will be better to show `prior` alongside with `observed`. `Inferred` is the result of SMC simulations. How can we interpret influence of `observed` to change `prior` in this picture?

reubenharry:
> I thought it will be better to show `prior` alongside with `observed`. `Inferred` is the result of SMC simulations. How can we interpret influence of `observed` to change `prior` in this picture?

This is an example of a Bayes filter, I believe. "Prior" represents a prior, not over individual points, but over an entire trajectory. "Inferred" represents the (estimated) posterior over this trajectory, given the observation of an "observed" trajectory. Let me know if that's useful, if not I can say more.

reubenharry:
I think explaining the implementation of SMC makes sense, but I do like that these tutorials can be read as user-facing documents too, for people who want to use SMC but not understand how it works.

mknorps:
OK. In that case I will not be putting a more description there. I linked Adam Scibor's [article](https://dl.acm.org/doi/pdf/10.1145/3236778).

reubenharry (approved):

PR #232, opened 2022-11-17
Add bus example and necessary Poisson pdf

idontgetoutmuch:
I took Sam Staton's bus example from https://www.cs.uoregon.edu/research/summerschool/summer19/topics.php#Staton and added it as an example. This required adding the Poisson pmf but I seem to have called it poissonPdf because that is what Sam did. Apart from naming, I think this is uncontroversial.

### DIFF
--- a/src/Control/Monad/Bayes/Traced/Static.hs
+++ b/src/Control/Monad/Bayes/Traced/Static.hs
@@ -83,7 +83,6 @@ mhStep (Traced m d) = Traced m d'
     d' = d >>= mhTransFree m
 
 -- $setup
--- >>> let z = 23 :: Int
 -- >>> import Control.Monad.Bayes.Class
 -- >>> import Control.Monad.Bayes.Sampler.Strict
 -- >>> import Control.Monad.Bayes.Weighted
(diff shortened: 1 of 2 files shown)


src/Control/Monad/Bayes/Traced/Static.hs
```diff
@@ -82,8 +82,26 @@ mhStep (Traced m d) = Traced m d'
...
     d' = d >>= mhTransFree m
 
+-- $setup
+-- >>> import Control.Monad.Bayes.Class
+-- >>> import Control.Monad.Bayes.Sampler.Strict
+-- >>> import Control.Monad.Bayes.Weighted
+
 -- | Full run of the Trace Metropolis-Hastings algorithm with a specified
 -- number of steps. Newest samples are at the head of the list.
+--
```
mknorps, line 92: Could you add a description of Sam Staton's bus example? It will be easier to understand with the context.

mknorps (changes requested):

src/Control/Monad/Bayes/Traced/Static.hs
idontgetoutmuch: How's this?

mknorps: perfect!

mknorps (approved):

PR #234, opened 2022-11-24
Allow GHC 9.4

turion:
I tested with GHC 9.4 and it compiles and tests fine.


mknorps (approved):
Tests done for GHC: 9.0.2, 9.2.4, 9.4.2:
```
nix-shell -p haskell.compiler.ghcXYZ
cabal build
cabal test
```
where XYZ is a GHC version.
All passed.
Also tested few notebooks and they work well.

PR #236, opened 2022-12-22
New docs

reubenharry:
Mkdocs. Covers both docs and website. Looks nicer. Still on netlify.


reubenharry:
@mknorps @idontgetoutmuch Here are what the new docs look like: https://monad-bayes.netlify.app/ The motivation was that currently I'm hosting two separate sites, one for the docs and one for the notebooks, and this both looks cleaner and is easier to maintain. I think this repo is set up such that I can't merge this in myself, so if one of you could, that would be great, and then I can delete the old websites.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,17 @@
+name: ci 
+on:
+  push:
+    branches:
+      - newdocs
```
mknorps, line 5: Change trigger to push on master and add cache.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,17 @@
...
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: 3.x
+      - run: pip install mkdocs-material 
+      - run: mkdocs gh-deploy --force
```
mknorps: Why do we need `--force` here?

mkdocs.yml
```diff
@@ -0,0 +1,71 @@
+site_name: Probabilistic Programming in Haskell
+theme:
+  name: material
+  features:
+    # - announce.dismiss
```
mknorps: Are the commented features still needed? If not, please remove.

mkdocs.yml
```diff
@@ -0,0 +1,71 @@
...
+    - navigation.sections
+    # - navigation.tabs
+    # - navigation.tabs.sticky
+    - navigation.top
+    - navigation.tracking
+    - search.highlight
+    - search.share
+    - search.suggest
+    - toc.follow
+  palette:
```
mknorps: Nice! I wonder if this will work with bright and dark mode of the users' browsers.

runtime.txt
```diff
@@ -0,0 +1 @@
+3.8
```
mknorps: It would be better to use Python 3.10 or 3.11. The maintanance period for them will be longer.

docs/usage.md
```diff
@@ -654,23 +654,6 @@ model = do
 
 is really an unnormalized measure, rather than a probability distribution. `normalize` views it as of type `Weighted Integrator Double`, which is isomorphic to `(Double -> (Double, Log Double) -> Double)`. This can be used to compute the normalization constant, and divide the integrator's output by it, all within `Integrator`.
```
mknorps: Why you do not want to include "Quadrature" in this version of the documentation?

mknorps (changes requested):
The site looks elegant! It has better user experience and you modernized it well. I like it very much.

Apart from the comments within the code, I have few comments and suggested changes regarding the PR:
- in `docs/images` plot.pgn and plot copy.pgn are duplicates. Remove the copy.
- How does `tintin` work here? I see a `.tintin.yml` file but no reference to it in teh code and requirements.
- Do we still need `monad-bayes-site` folder?
- For the structure of the project it would be better if the code to generate documentation was in a separate folder. 
- is `monad-bayes-site.cabal` and `stack.yaml` within `docs` still needed?

docs/usage.md
reubenharry: It was duplicated.

runtime.txt
reubenharry: Done.

.github/workflows/ci.yml
reubenharry: Removed.

.github/workflows/ci.yml
reubenharry: Wasn't sure how to add cache, but added master.

reubenharry:
For some reason, the --force and Python 3.8 were both necessary for CI and netlify to work respectively, but I changed the other things.

.github/workflows/ci.yml
mknorps: This should work:
```
      - name: Setup caching
        uses: actions/cache@v3
        with:
          key: ${{ github.sha }}
          path: .cache
```
Check out: https://gaseri.org/en/blog/2022-11-01-publishing-material-for-mkdocs-website-to-github-pages-using-custom-actions-workflow/

mknorps:
> For some reason, the --force and Python 3.8 were both necessary for CI and netlify to work respectively, but I changed the other things.

That is inconvenient, but we can leave with this for now.

And what about:
> Do we still need monad-bayes-site folder?
> For the structure of the project it would be better if the code to generate documentation was in a separate folder.

Is also mathjax no longer needed?

reubenharry:
> And what about:
> 
> > Do we still need monad-bayes-site folder?

I thought I removed it. I don't see it in the repo locally.

> > For the structure of the project it would be better if the code to generate documentation was in a separate folder.
> 

Now done.

> Is also mathjax no longer needed?

Re-added.

mknorps (changes requested):
Thank you! The code structure looks better :)
One last thing - remove mkdocs.yaml from the main directory.

If you want to experiment a bit with caching, as I suggested in [this comment ](https://github.com/tweag/monad-bayes/pull/236/files#r1059271173) it would be great, but if you do not have time for this now, please create an issue for later not to forget about it.

reubenharry:
OK, removed, and issue created.

reubenharry:
Hm, netlify works fine, but something with CI fails. I'm a bit stumped. Any ideas?

reubenharry:
Oh right, it's failing for the old site. Let me remove it...

reubenharry:
Now passes CI, should be ready to merge.

mknorps (approved):
It looks OK now. Thank you for improving the documentation!

I created https://github.com/tweag/monad-bayes/issues/239 to take care of the developer's environment regarding developing documentation.

Issue #241, opened 2023-01-10
Resampling a Population does not normalize


turion:
**Describe the bug**
When applying a resampler, the total evidence stays the same.

**To Reproduce**

```
ghci> import Control.Monad.Bayes.Population
ghci> import Control.Monad.Bayes.Sampler.Strict
ghci> let model = spawn 1000 >> normal 0 1 >>= (condition  . (>= 0))
ghci> sampleIOfixed  $ evidence model
0.48900000000000027
ghci> sampleIOfixed  $ evidence $ resampleStratified model
0.48900000000000027
```

**Expected behavior**

I'm not sure whether this is intended. It seems like it, given this line: https://github.com/tweag/monad-bayes/blob/110e040099d7f1dc1c5f0bd6170fed759897a5b6/src/Control/Monad/Bayes/Population.hs#L116

Extra care is taken to make sure that the total mass from before (`z`) is restored, and no accidental normalization is introduced.

There used to be a function `normalize`, which was removed in 771ce2ece83f0de7cc0d592bec6dd6e660024ddc by @reubenharry, as part of https://github.com/tweag/monad-bayes/pull/142. It was part of the 0.1.1.0 release: https://hackage.haskell.org/package/monad-bayes-0.1.1.0/docs/Control-Monad-Bayes-Population.html#v:normalize

Its haddocks read:

 > Normalizes the weights in the population so that their sum is 1. This transformation introduces bias.

As of `master`, there are other ways to "normalize" the distribution, but they all require `MonadFactor` on the inner monad. This is sometimes not feasible. For example when doing particle filtering (see https://github.com/turion/dunai-bayes/), one needs to condition and resample on every step of the time series. Then the probability masses of the individual particles drop at an exponential rate, no matter how much we condition in the inner monad. Very soon, they shrink below `2^(-64)` and cannot be resolved by a `Double` anymore. In these situations, I would have believed that it makes sense to normalize, so the numbers stay in a region we can calculate with.

What I don't understand are the comments about "bias". In 1d5521840e9081603bc760621af3d9bc2685da9c it is commented by @adscib that `normalize` indroduces bias, and I don't understand what that means. In particular I don't understand whether normalizing at every time step in a particle filter then also introduces bias. If yes, how do other implementations of particle filters deal with this situation?

reubenharry:
[...]

PR #243, opened 2023-01-11
Re-introduce normalize

idontgetoutmuch:


turion:
Intends to close https://github.com/tweag/monad-bayes/issues/241

src/Control/Monad/Bayes/Population.hs
```diff
@@ -274,3 +276,7 @@ hoist ::
   Population m a ->
   Population n a
 hoist f = fromWeightedList . f . population
+
+-- | Normalizes the weights in the population so that their sum is 1.
+normalize :: (Monad m) => Population m a -> Population m a
+normalize = hoist unweighted . extractEvidence
```
turion, line 282: `unweighted` used to be called `prior` https://hackage.haskell.org/package/monad-bayes-0.1.1.0/docs/src/Control.Monad.Bayes.Population.html#normalize, so I guess this is the same implementation as before. Shouldn't we restore the comment about bias here?

turion:
In `test/TestInference.hs`, you should hide the import `normalize` from `Population`

Issue #245, opened 2023-01-17
ListT is deprecated


turion:
# Situation

`ListT` is deprecated since a long time, and has been removed from major packages:

https://github.com/haskell/mtl/issues/88
https://hackage.haskell.org/package/transformers-0.6.0.0/changelog

The reason it is deprecated is because for a noncommutative monad `m`, `ListT m` famously is not a monad. `m` needs to be commutative for it to be a monad, but that restriction is not enforced anywhere, so it was better to drop the whole construction.

## Situation for `monad-bayes`

This also applies to randomness monads like we use here. Some monads like `Weighted` are commutative, but others really aren't. For example:

```haskell
mv1 = do
  x <- normal 0 1
  y <- normal 0 1
  return (x, y)
mv2 = do
  y <- normal 0 1
  x <- normal 0 1
  return (x, y)
```
 While `mv1` and `mv2` define the same probability distribution mathematically, they are not the same value. When we sample from them with the same random seed, they will produce different results. Sampling is not commutative.

Side note: If we are only interested in statistics calculated from the samples in some kind of limit where we sample arbitrarily often, then we might not care about the difference. But if, for example, we want to reproduce some random sample path, the laws are broken.

### Bit rot

The main problem arising from this situation is that `monad-bayes` slowly bit-rots because we cannot update dependencies. We have to restrict version bounds to `mtl < 2.3 && transformers < 0.6`, which will eventually keep `monad-bayes` out of stackage and make it harder for everyone to use it.

### Broken laws
The side problem is also that right now we have a few functions that expose the breakage of monad laws: We have functions like [`runPopulation`](https://hackage.haskell.org/package/monad-bayes-1.1.0/docs/Control-Monad-Bayes-Population.html#v:runPopulation) that expose the list order, so the broken monad laws can be observed.

### Performance
I haven't measured this yet, but I'll venture the guess that `ListT` is not very fast because it uses lists. Iterating over long lists of particles will incur some performance penalty.
# What could be done

### Bit rot

We could ship our own `ListT` explicitly and thus remove the restrictive upper bounds on `mtl`.

### Broken laws

The concrete order of the list elements must never be able to be observed. If we can only, e.g. sample a list element, or only retrieve all list elements in a fixed order, then this is fine. For example, these type signatures are ok:

```haskell
-- Sorts e.g. by probability, then by a
runPopulation :: Ord a => Population m a -> m [(a, Log Double)]
-- The ordering information is lost in this step
[...]

PR #253, opened 2023-02-07
Free list transformer to replace deprecated ListT

turion:
[samples_free_listt.pdf](https://github.com/tweag/monad-bayes/files/10677295/samples_free_listt.pdf)
[samples_master.pdf](https://github.com/tweag/monad-bayes/files/10677296/samples_master.pdf)

Removes the dependency on the deprecated `ListT` in favour of a free monad construction. In fact _improves_ benchmarks significantly!

* [x] Merge #255 first
* [x] Merge #256 first!

### DIFF
--- a/src/Control/Monad/Bayes/Population.hs
+++ b/src/Control/Monad/Bayes/Population.hs
@@ -56,7 +56,7 @@ import Control.Monad.Bayes.Weighted
 import Control.Monad.Bayes.Weighted qualified as Weighted
 import Control.Monad.IO.Class
 import Control.Monad.Trans
-import Control.Monad.Trans.Free.Ap
+import Control.Monad.Trans.Free
 import Data.List (unfoldr)
 import Data.List qualified
 import Data.Maybe (catMaybes)
--- a/src/Control/Monad/Bayes/Population.hs
+++ b/src/Control/Monad/Bayes/Population.hs
@@ -40,7 +40,6 @@ module Control.Monad.Bayes.Population
 where
 
 import Control.Applicative (Alternative)
-import Control.Applicative.List qualified as ApplicativeListT
 import Control.Arrow (second)
 import Control.Monad (MonadPlus, replicateM)
 import Control.Monad.Bayes.Class
@@ -49,6 +48,7 @@ import Control.Monad.Bayes.Class
     MonadMeasure,
     factor,
   )
+import Control.Monad.Bayes.Population.Applicative qualified as Applicative
 import Control.Monad.Bayes.Weighted
   ( WeightedT,
     applyWeight,
@@ -70,6 +70,11 @@ import Numeric.Log qualified as Log
 import Prelude hiding (all, sum)
 
 -- | A collection of weighted samples, or particles.
+--
+-- This monad transformer is internally represented as a free monad,
+-- which means that each layer of its computation contains a collection of weighted samples.
+-- These can be flattened with 'flatten',
+-- but the result is not a monad anymore.
 newtype PopulationT m a = PopulationT {getPopulationT :: WeightedT (FreeT [] m) a}
   deriving newtype (Functor, Applicative, Alternative, Monad, MonadIO, MonadPlus, MonadDistribution, MonadFactor, MonadMeasure)
 
@@ -278,12 +283,12 @@ hoist ::
   PopulationT n a
 hoist f = PopulationT . Weighted.hoist (hoistFreeT f) . getPopulationT
 
--- | Flatten all layers of the free structure
-flatten :: (Monad m) => PopulationT m a -> ApplicativeListT.PopulationT m a
-flatten = ApplicativeListT.fromWeightedList . runPopulationT
+-- | Flatten all layers of the free structure.
+flatten :: (Monad m) => PopulationT m a -> Applicative.PopulationT m a
+flatten = Applicative.fromWeightedList . runPopulationT
 
 -- | Create a population from a single layer of branching computations.
 --
 -- Similar to 'fromWeightedListT'.
-single :: (Monad m) => ApplicativeListT.PopulationT m a -> PopulationT m a
-single = fromWeightedList . ApplicativeListT.runPopulationT
+single :: (Monad m) => Applicative.PopulationT m a -> PopulationT m a
+single = fromWeightedList . Applicative.runPopulationT
--- a/src/Control/Monad/Bayes/Population.hs
+++ b/src/Control/Monad/Bayes/Population.hs
@@ -34,6 +34,7 @@ module Control.Monad.Bayes.Population
     collapse,
     popAvg,
     withParticles,
+    flatten,
   )
 where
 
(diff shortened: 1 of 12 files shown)


turion:
I observed runtimes to go down by 30%, and never going up in comparison to `master`. This is probably because the free monad over an applicative only adds free layers whenever it really needs, but not for `fmap` and `<*>`. I didn't test whether a free monad over the list functor gives a similar speedup, but it has no advantage I'm aware of, so I didn't 

Adding `ApplicativeDo` globally also had a slight positive effect, but that's a separate issue.

src/Control/Monad/Bayes/Population.hs
```diff
@@ -269,7 +271,7 @@ popAvg f p = do
 
 -- | Applies a transformation to the inner monad.
 hoist ::
-  Monad n =>
```
reubenharry: Are you still planning to replace this with MMorph?

reubenharry (commented):
This is amazing! My one request would be an update to the documentation here: https://monad-bayes.netlify.app/usage/#population You can update this by changing the relevant markdown file in the docs. Just a quick explanation of the motivation of the new representation, where it comes from, maybe the definition too. I can also have a go at this, but would be nice to check if it's straightforward for you to update the docs.

src/Control/Monad/Bayes/Population.hs
turion: Yes, eventually. [`hoist`](https://hackage.haskell.org/package/mmorph-1.2.0/docs/Control-Monad-Morph.html#v:hoist) from `mmorph` brings the `Monad m` constraint, so this change should be fine. It's the `Monad n` constraint that is tricky.

turion:
> My one request would be an update to the documentation here: https://monad-bayes.netlify.app/usage/#population 

Yes, on it.

> but would be nice to check if it's straightforward for you to update the docs.

Very good point.

I have trouble understanding the current documentation:

```
Population m a ~ m [Log Double -> (a, Log Double)]
```
Shouldn't it be:
```
Population m a ~ m [(a, Log Double)]
```

reubenharry:
```hs 
newtype Population m a = Population (Weighted (ListT m) a)
```

`Weighted` is the `State` monad. (By the way, it's possible that one of the optimized Writer or accum monads could be used here, but that's for a separate issue).

ohad:
Nice!
Since this implementation is computing things differently, not just faster, we should at least try to check that it's producing the correct results, not just producing results faster.

turion:
> Weighted is the State monad. (By the way, it's possible that one of the optimized Writer or accum monads could be used here, but that's for a separate issue).

Ah, true. I guess it's implemented as `State` for efficiency reasons, but its interface only allows `Writer` operations. Ironically in the `ListT` situation that's an important distinction because `State` is not commutative, but `Writer` for a commutative monoid like `Log Double` is.

I wonder what we should do documentation-wise. I'd find it simpler to explain that `Weighted` is a `Writer`, so we can't arbitrarily change the weight of a path, but only adding to its log. But that's a separate issue.

ohad:
We should understand what the proposed change is doing a bit more before merging. I wrote it in issue #245, but it also belongs here.

The [paper](https://arxiv.org/pdf/1711.03219.pdf) and its [follow-up](https://dl.acm.org/doi/10.1145/3236778) used the monad-structure transformer `ListT` because the computation it produces makes the correct samples and in the correct order. @adscib (and later the Tweag team) have tested that the resulting inference algorithms produces the expected distributions when run on models, and the [paper](https://arxiv.org/pdf/1711.03219.pdf) proves the algorithms, when working on these data structures, are correct. 

In general, free monads may do something different, for example, suspend random samples that some of the monad-bayes combinators do in crucial places. This might mean, for example, that using the same combinations of monad-bayes's building blocks to implement algorithms such as smc, smc^2 etc. are not longer correct.

I suggest we first understand better how this proposed change affects the existing inference algorithms, not just in runtime, but in the actual samples they compute, both by testing and using an accompanying formalisation.

src/Control/Monad/Bayes/Population.hs
```diff
@@ -269,8 +273,12 @@ popAvg f p = do
...
+  (Monad m, Monad n) =>
   (forall x. m x -> n x) ->
   Population m a ->
   Population n a
 hoist f = fromWeightedList . f . population
+
+-- | Flatten all layers of the free structure
+flatten :: Monad m => Population m a -> Population m a
```
turion: There is a design space here. Instead of making this an internal operation of `Population` which is invisible to the type checker, we could also have a second newtype:

```haskell
newtype FlatPopulation m a = FlatPopulation (m [a])
  deriving (Functor, Applicative) -- Note, no Monad
```
(Naming is up for grabs, we could also call the other thing `FreePopulation`.)
Then there is a function `Population m a -> FlatPopulation m a`. This has the advantage that we can specify in the different algorithms like `rmsmc`, `resampleGeneric` etc. when we need the structure flattened. That way we get a better grip on not making any mistakes with the order.

reubenharry:
> Ah, true. I guess it's implemented as `State` for efficiency reasons, but its interface only allows `Writer` operations. 

That's right. I'm still surprised that there isn't a Writer implementation that is suitably efficient (or even just uses State under the hood), but I don't know enough about performance to solve that problem. (I tried the continuized Writer, and I recall seeing mild speedups or at least, no slowdowns, although I might be misremembering).

> I wonder what we should do documentation-wise. I'd find it simpler to explain that `Weighted` is a `Writer`, so we can't arbitrarily change the weight of a path, but only adding to its log. But that's a separate issue.

I agree in theory, but I think this would just cause extra confusion when anyone looked at the code and saw State, not Writer.

turion:
> I suggest we first understand better how this proposed change affects the existing inference algorithms, not just in runtime, but in the actual samples they compute, both by testing and using an accompanying formalisation.

I tried to go for a conservative approach now that keeps an `Applicative` `PopulationT` transformer which has the same semantics as before, and is used in all places where needed. Unfortunately it isn't clear to me where to appropriately flatten the free population trees to make the SMC2 algorithm work. I tried all places I could think of, but ultimately didn't succeed in fixing the corresponding test. (On the other hand, I don't know whether SMC2 was correct to begin with.)

One learning is definitely that using an unlawful monad is a maintenance nightmare. Replacing it with a lawful one and then trying to find all the places where the particular flattening behaviour was intended and where it was accidental is like looking for a needle in a haystack. If someone else wants to give this a go, I'm happy to hand over and advise, but I've done enough work on this now for free.

Another thought I had while working on this issue: `SequentialT` is used pretty much like a free monad transformer. But the disadvantage is that it is defined using a library that is not very well known and popular. Possibly, we should do away with it completely and just use free monads instead. This might clarify the `PopulationT` story.

reubenharry:
I also had that thought re. SequentialT, and attempted to replace with FreeT and a catamorphism, but discovered that my understanding was not good, because the "folding" in the coroutine operations was quite different to what I'd imagined. You might have more luck though.

PR #255, opened 2023-02-08
Warnings as errors

turion:
I think it would be a good idea to catch GHC warnings early on. This PR makes sure CI fails when there is a GHC warning.

CC @idontgetoutmuch


turion:
Naturally, the build fails because there are warnings in the current tree.

idontgetoutmuch (approved):

idontgetoutmuch:
@turion can you fix the warning and then resubmit?

turion:
@idontgetoutmuch can we not merge yet? It would be nice to merge without squashing.

idontgetoutmuch:
Great! Thanks for doing this.

Issue #191, opened 2022-09-13
Generalize from Double


reubenharry:
Use type families in type classes to move from Double to an instance dependent number type. (Seems to work, just needs to be polished.)

idontgetoutmuch:
I also tried to do this but without type families but have so far not succeeded. Here's a gist containing everything necessary in one file: https://github.com/tweag/monad-bayes/blob/Generalising-the-sampler/test/gist.hs.

I did also try your (@reubenharry) type family approach but was foxed by the type error given by this function

```
mhTrans' :: (StatefulGen g m, MonadReader g m) =>
             MonadDistribution m => Weighted (Density Identity) a -> Trace a -> m (Trace a)
mhTrans' m = mhTrans (hoistW (hoistD (return . runIdentity)) m)
```

I notice that you (@reubenharry) also didn't complete this: https://github.com/tweag/monad-bayes/blob/generalizeFromDouble2/src/Control/Monad/Bayes/Traced/Common.hs#L98.

```
-- | A variant of 'mhTrans' with an external sampling monad.
mhTrans' :: MonadSample m => Weighted (Density Identity) a -> Trace (Real m) a -> m (Trace (Real m) a)
mhTrans' m = undefined -- mhTrans (Weighted.hoist (FreeSampler.hoist (return . runIdentity)) m)
```

If either you (@reubenharry) or @turion had any ideas, they would be gratefully received.

PS I added in some code to do AD.

PPS My plan is to implement https://arxiv.org/pdf/2106.10238.pdf and then https://arxiv.org/pdf/2211.01100.pdf.

PR #256, opened 2023-02-08
Test some benchmark configurations against fixtures

turion:
@ohad suggested to test inference algorithms for the samples they produce in https://github.com/tweag/monad-bayes/issues/245#issuecomment-1422385283. I think this is a good idea independently of the issue discussed there. So I implemented a few fixture tests for the three algorithms and the three models implemented in the benchmarks.


turion:
Are all inference algorithms covered by such a test now, and with sufficiently complex models?

reubenharry:
To make sure I understand (I'm probably confused): is the idea to ensure that the code always produces the same samples given the same random seed? If so, is this a property we always want? For example, what if we change the rng? Presumably the correctness of monad-bayes does not hinge on using the Mersenne twister rng (and indeed, Dominic and I made monad-bayes parametric on the choice of rng) but would this cause the tests to fail?

On this front, note that to be really safe, we might also want to check that the version of monad-bayes from before Spring 2022 (which we can be reasonably confident was in-sync with the proven correct implementation of the paper) agrees on these tests with the current version. I think I avoided making semantically impactful changes, but it's always subtle.

turion:
> To make sure I understand (I'm probably confused): is the idea to ensure that the code always produces the same samples given the same random seed? If so, is this a property we always want?

Yes, for changes that are not supposed to change the semantics.

> For example, what if we change the rng?

Then we should isolate this change in a single commit and update the fixtures there.

> Presumably the correctness of monad-bayes does not hinge on using the Mersenne twister rng (and indeed, Dominic and I made monad-bayes parametric on the choice of rng) but would this cause the tests to fail?

Yes it might. As would changing the random seed hardcoded in `sampleIOFixed`. But that's ok, running the tests updates the fixtures.

> On this front, note that to be really safe, we might also want to check that the version of monad-bayes from before Spring 2022 (which we can be reasonably confident was in-sync with the proven correct implementation of the paper) agrees on these tests with the current version. I think I avoided making semantically impactful changes, but it's always subtle.

Ok, I can try that if it's possible to backport the tests to there.

turion:
The tests are failing on darwin apparently because the precision of the floating point numbers used there is ever so slightly different. That's really annoying. Some options:

* [x] Don't fixture test on darwin
* [ ] Find out what is different, and use a type with system-independent precision. Might need to solve #191 first.

reubenharry:
My feeling would be that for now it's worth just ignoring Darwin and merging, especially since 191 risks changing the fixtures.

turion:
Ok, I've tried to disable tests on MacOS, let's see whether that works.

test/fixtures/HMM10-MH.txt
```diff
@@ -0,0 +1 @@
..."[1,2,1,1,1,1,1,2,2,0]","[2,2,1,1,1,1,1,2,2,0]","[2,2,1,1,1,2,1,1,2,0]","[2,2,1,1,1,2,1,1,2,0]","[2,2,1,1,1,2,1,1,2,0]","[2,2,1,1,1,2,1,1,2,0]","[2,2,1,1,1,2,1,1,2,0]","[2,1,2,1,1,2,1,1,2,0]","[2,1,2,1,1,2,1,1,2,0]","[2,1,2,1,1,2,1,1,2,0]","[2,1,2,1,1,2,1,1,2,0]","[1,1,2,1,1,2,1,1,2,0]","[1,2,1,1,1,2,1,1,2,0]","[1,2,1,1,1,2,1,1,2,0]","[1,1,1,1,1,2,1,1,2,0]","[1,1,1,1,1,2,1,1,2,0]","[1,1,1,1,1,2,1,1,2,0]","[1,1,1,1,1,0,1,1,2,0]","[1,1,1,1,1,0,1,1,2,0]","[1,1,1,1,1,0,1,1,2,0]","[1,1,1,1,1,0,1,1,2,0]","[1,1,1,1,1,0,1,1,2,0]","[1,1,1,1,1,0,1,1,2,0]","[1,1,1,1,1,0,1,1,2,0]","[1,1,1,1,1,0,1,1,2,0]","[1,1,1,1,1,0,1,1,2,0]","[1,1,1,1,1,0,1,1,2,0]","[1,1,1,1,1,0,1,1,2,0]","[1,1,1,1,1,0,1,1,2,0]","[1,1,1,1,1,0,1,1,2,0]","[1,1,1,1,1,0,1,1,2,0]","[1,1,1,1,1,0,1,1,2,0]","[1,1,1,1,1,0,1,1,2,0]","[1,1,1,1,1,0,1,1,2,0]","[1,1,1,1,1,0,1,1,2,0]","[1,1,1,1,1,0,1,1,2,0]","[1,1,1,1,1,0,1,1,2,0]","[1,1,1,1,1,0,1,1,2,0]","[1,1,1,1,1,0,1,1,2,0]","[1,1,1,1,1,0,1,1,2,0]","[1,1,1,1,1,0,1,1,2,0]","[1,1,1,1,1,0,1,1,2,0]","[0,1,1,1,1,0,1,1,2,0]","[0,1,1,1,1,0,1,1,2,0]","[0,1,1,1,1,0,1,1,2,0]","[0,1,1,1,1,0,1,1,2,0]","[0,1,1,1,1,0,1,1,2,0]","[0,1,1,1,1,0,1,1,2,0]","[0,1,1,1,1,0,1,1,2,0]","[0,1,1,1,1,0,1,1,2,0]"]
```
turion, line 1: I'm a bit surprised by this example data because it stays largely constant. I'm still trying to understand what that means.

turion:
> My feeling would be that for now it's worth just ignoring Darwin and merging,

I'll interpret this as approval ;)

test/TestSSMFixtures.hs
```diff
@@ -0,0 +1,80 @@
...
+
+fixtureToFilename :: Alg -> FilePath
+fixtureToFilename alg = "test/fixtures/SSM-" ++ show alg ++ ".txt"
+
+type SSMData = [Double]
+
+t :: Int
+t = 5
+
+-- FIXME refactor such that it can be reused in ssm benchmark
```
turion: Still to be resolved

turion:
@reubenharry can you review?

turion:
@reubenharry @idontgetoutmuch can you review this PR?

reubenharry (approved):
Again, seems good.

PR #266, opened 2023-05-10
Allow different versions of GHC on CI

turion:
This adds a list of GHC versions in `flake.nix` which can be extended. Each of these versions is used to build and test the package.


turion:
Right now every PR fails because of formatting for some reason. I suspect some version of the formatter was increased without running CI completely.

turion:
I'll try to fix in https://github.com/tweag/monad-bayes/pull/269 first.

turion:
@idontgetoutmuch If you're ok with this we can at least claim to support two GHC versions now :)

idontgetoutmuch (approved):
Great - thanks for doing this :-)

PR #267, opened 2023-05-10
Allow 9.6

turion:
Follow-up to https://github.com/tweag/monad-bayes/pull/266 (should be merged first).

* [x] Clean up git history

### DIFF
--- a/benchmark/Single.hs
+++ b/benchmark/Single.hs
@@ -1,12 +1,12 @@
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE ImportQualifiedPost #-}
 
+import Control.Applicative (Applicative (..))
 import Control.Monad.Bayes.Sampler.Strict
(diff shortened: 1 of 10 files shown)


turion:
It seems that GHC 9.6 is nigh impossible because ListT is really, really gone from the ecosystem. I'll aim at 9.4 first and see whether we can get #245 resolved first.

reubenharry:
Surely the ListT code we need is fairly minimal though? Can't we just put it directly in the repo?

turion:
If there is no great hurry to support GHC 9.6 right away I'd rather go with the clean solution from the start.

turion:
> I'll aim at 9.4 first

This is https://github.com/tweag/monad-bayes/pull/271

reubenharry (approved):
Nothing objectionable here from my perspective - glad to see 9.6 here though!

PR #268, opened 2023-05-10
Add effective sample size to Population

turion:
I recently read about improving existing resampling methods by only resampling when the effective sample size is small.

* [x] Complete fixture test MR #256

### DIFF
--- a/src/Control/Monad/Bayes/Population.hs
+++ b/src/Control/Monad/Bayes/Population.hs
@@ -27,6 +27,8 @@ module Control.Monad.Bayes.Population
     resampleSystematic,
     stratified,
     resampleStratified,
+    onlyBelowEffectiveSampleSize,
+    effectiveSampleSize,
     extractEvidence,
     pushEvidence,
     proper,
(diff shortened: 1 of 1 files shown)


reubenharry (approved):
I assume this preserves the unbiased nature of SMC (since no resampling is unbiased, and full resampling is unbiased) - if not it should be commented.

reubenharry (approved):
I assume this preserves the unbiased nature of SMC (since no resampling is unbiased, and full resampling is unbiased) - if not it should be commented.

turion:
> I assume this preserves the unbiased nature of SMC (since no resampling is unbiased, and full resampling is unbiased) - if not it should be commented.

I'm not sure. Is there a resource to read up about biases introduced by resampling? I didn't read anything in Murphy's book about this not being unbiased. I would have thought that if the procedure at each timestamp is not biased, the whole process is not biased, but I can't prove it.

turion:
Also, I want to add a fixture test and look at the values a bit before merging.

reubenharry:
I don't really know, but yeah, I also agree with that reasoning.

turion:
I managed to produce a crash using this function in `rhine-bayes`, but I haven't found the reason yet.

turion:
> I managed to produce a crash using this function in `rhine-bayes`, but I haven't found the reason yet.

The crash was:

```
rhine-bayes-gloss: Statistics.Distribution.Normal.normalDistr: standard deviation must be positive. Got NaN
CallStack (from HasCallStack):
  error, called at ./Statistics/Distribution/Normal.hs:115:27 in statistics-0.16.2.0-ca3954dbbfc16db2d02d8c716cbf789c747e026d3f9db91380356ec3f9721ef8:Statistics.Distribution.Normal
Segmentation fault (Speicherabzug geschrieben)
```

I could not manage to reproduce it again. In subsequent runs, it seemed to me that one could definitely speed up performance, while keeping the same variance.

src/Control/Monad/Bayes/Population.hs
```diff
@@ -210,6 +212,31 @@ resampleMultinomial ::
...
+  MonadDistribution m =>
+  -- | The threshold under which the effective sample size must fall before the resampler is used.
+  --   For example, this may be half of the number of particles.
+  Double ->
+  -- | The resampler to user under the threshold
+  (MonadDistribution m => Population m a -> Population m a) ->
+  -- | The new resampler
+  (Population m a -> Population m a)
+onlyBelowEffectiveSampleSize threshold resampler pop = do
+  ess <- lift $ effectiveSampleSize pop
```
turion: I think this is wrong because it will execute the `m` effects of `pop` again.

turion: It might be better to have a function `withEffectiveSampleSize :: Functor m => Population m a -> m (a, Double)`

src/Control/Monad/Bayes/Population.hs
```diff
@@ -206,6 +208,31 @@ resampleMultinomial ::
   PopulationT m a
 resampleMultinomial = resampleGeneric multinomial
 
+-- | Only use the given resampler when the effective sample size is below a certain threshold
+onlyBelowEffectiveSampleSize ::
```
turion, line 254: I should have one fixture test and a unit test where this is used.

PR #257, opened 2023-02-09
Simplify lock file

turion:
I removed some duplicate inputs.

PR #269, opened 2023-05-10
Fix formatter

turion:


turion:
Probably complete #257 first, maybe this PR isn't even needed anymore.

PR #271, opened 2023-05-16
Enable GHC 9.4

turion:
Part of https://github.com/tweag/monad-bayes/pull/267

### DIFF
--- a/monad-bayes.cabal
+++ b/monad-bayes.cabal
@@ -7,7 +7,7 @@ copyright:          2015-2020 Adam Scibior
 maintainer:         dominic.steinitz@tweag.io
 author:             Adam Scibior <adscib@gmail.com>
 stability:          experimental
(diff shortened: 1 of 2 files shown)


monad-bayes.cabal
```diff
@@ -64,7 +64,7 @@ library
   other-modules:      Control.Monad.Bayes.Traced.Common
   default-language:   Haskell2010
   build-depends:
-      base             >=4.11    && <4.18
```
turion: This is not necessary

reubenharry (approved):

turion:
It works pretty much out of the box. But we should be aware that this adds time to CI. It becomes more important to fix the Cachix setup.

turion:
> But we should be aware that this adds time to CI. It becomes more important to fix the Cachix setup.

See:
image

It seems like it builds a lot from scratch, and every GHC adds something like half an hour of CI time. This will be a major inconvenience.

PR #272, opened 2023-05-16
Add newtype to lift classes to monad transformers

turion:
With this PR, it should be possible to derive the `MonadMeasure` class easily for any monad transformer.

### DIFF
--- a/src/Control/Monad/Bayes/Class.hs
+++ b/src/Control/Monad/Bayes/Class.hs
@@ -345,69 +345,45 @@ histogramToList = H.asList
 ----------------------------------------------------------------------------
 -- Instances that lift probabilistic effects to standard tranformers.
 
-instance MonadDistribution m => MonadDistribution (IdentityT m) where
-  random = lift random
-  bernoulli = lift . bernoulli
+deriving via (MonadMeasureTrans IdentityT m) instance MonadDistribution m => MonadDistribution (IdentityT m)
 
-instance MonadFactor m => MonadFactor (IdentityT m) where
-  score = lift . score
+deriving via (MonadMeasureTrans IdentityT m) instance MonadFactor m => MonadFactor (IdentityT m)
 
 instance MonadMeasure m => MonadMeasure (IdentityT m)
 
-instance MonadDistribution m => MonadDistribution (ExceptT e m) where
-  random = lift random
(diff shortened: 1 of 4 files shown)


src/Control/Monad/Bayes/Class.hs
```diff
@@ -342,68 +345,82 @@ histogramToList = H.asList
...
-  score = lift . score
+deriving via (MonadMeasureTrans (WriterT w) m) instance (Monoid w, MonadFactor m) => MonadFactor (WriterT w m)
 
 instance (Monoid w, MonadMeasure m) => MonadMeasure (WriterT w m)
 
-instance MonadDistribution m => MonadDistribution (StateT s m) where
-  random = lift random
-  bernoulli = lift . bernoulli
-  categorical = lift . categorical
-  uniformD = lift . uniformD
```
reubenharry, line 386: My only note here would be that some of the previous lifting instances appear to have varied slightly. This actually may have been my doing, because I recall that working in a transformed Enumerator didn't lift `bernoulli`, so that discrete distributions got calculated via `random` and thus failed with `enumerate`.

src/Control/Monad/Bayes/Class.hs
```diff
@@ -342,68 +345,82 @@ histogramToList = H.asList
...
+-- (The final 'MonadMeasure' could also be derived `via`, but this isn't necessary because it doesn't contain any methods.)
+newtype MonadMeasureTrans (t :: (Type -> Type) -> Type -> Type) (m :: Type -> Type) a = MonadMeasureTrans {getMonadMeasureTrans :: t m a}
+  deriving (Functor, Applicative, Monad)
+
+instance MonadTrans t => MonadTrans (MonadMeasureTrans t) where
+  lift = MonadMeasureTrans . lift
+
+instance (MonadTrans t, MonadDistribution m, Monad (t m)) => MonadDistribution (MonadMeasureTrans t m) where
   random = lift random
+  uniform = (lift .) . uniform
```
reubenharry, line 411: Similarly here, there may have been a reason that many of the instances were limited to mostly just `lift random`. This doesn't change the denotational semantics, but may have efficiency implications, so make sure to benchmark, particularly PMMH and RMSMC.

reubenharry (approved):

PR #302, opened 2023-08-08
Convert notebooks to html automatically

turion:
This PR builds the `*.html` files from the notebooks and compares them to the checked in files. So if there are notebook changes, CI forces you to regenerate the `*.html`.


turion:
CC @reubenharry

kernels/haskell.nix
```diff
@@ -1,41 +1,41 @@
...
+      p.recursion-schemes
+      p.data-fix
+      p.free
+      p.comonad
+      p.adjunctions
+      p.distributive
+      p.vector
+      p.megaparsec
+      p.histogram-fill
+      # Strange build error
```
turion: I should try and fix this

turion: I couldn't. I tried overriding in any way I could think of, but it seems like `gloss` in `ihaskell` is broken because they pin `bytestring` to the wrong version.

turion: The only way I see for now is leaving this out, then one of the notebooks will not work out of the box.

turion:
I should make a branch on the main repo (and not my fork) so it caches artifacts.

turion:
Closed in favour of #304

PR #303, opened 2023-08-10
Allow vector-0.13

turion:
vector-0.13 is released, and it seems like `monad-bayes` works well with it. See https://github.com/NixOS/nixpkgs/pull/248112/files for a workaround that was necessary because of the restrictive upper bound.


turion:
CC @idontgetoutmuch @reubenharry can you review?

reubenharry (approved):

PR #322, opened 2023-10-13
Fix #320

### LINKED ISSUE #320: Lazy sampler is not a monad transformer, strict sampler misses MonadTrans instance
In https://github.com/tweag/monad-bayes/pull/295 I erroneously renamed [`Control.Monad.Bayes.Sampler.Lazy.Sampler`](https://hackage.haskell.org/package/monad-bayes-1.1.0/docs/Control-Monad-Bayes-Sampler-Lazy.html#t:Sampler)  to [`SamplerT`](https://hackage.haskell.org/package/monad-bayes-1.2.0/docs/Control-Monad-Bayes-Sampler-Lazy.html#t:SamplerT). Looking into this I found that:

* There is no lazy sampler transformer going with the lazy sampler
* The strict sampler is a transformer, but doesn't have the `MonadTrans` instance

This should be fixed

turion:
* Create lazy sampling monad transformer
* Rename Lazy.SamplerT to Lazy.Sampler and fix other naming issues
* Fix haddock for lazy sampler
* Add MonadTrans instances

CC @reubenharry @idontgetoutmuch can you review?

### DIFF
--- a/src/Control/Monad/Bayes/Inference/Lazy/MH.hs
+++ b/src/Control/Monad/Bayes/Inference/Lazy/MH.hs
@@ -7,10 +7,11 @@ module Control.Monad.Bayes.Inference.Lazy.MH where
 
 import Control.Monad.Bayes.Class (Log (ln))
 import Control.Monad.Bayes.Sampler.Lazy
-  ( SamplerT (runSamplerT),
+  ( Sampler,
     Tree (..),
     Trees (..),
     randomTree,
+    runSampler,
   )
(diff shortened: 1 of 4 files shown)


turion:
Ping @reubenharry

reubenharry (commented):
Yep, these changes all seem good to me. I added `independent` just for pedagogical purposes, but this way is fine too.

reubenharry (approved):

turion:
Fair enough, we can keep it.

Issue #294, opened 2023-07-17
Support GHC 9.6


turion:
Follow up to https://github.com/tweag/monad-bayes/issues/259.

Supporting GHC 9.6 is harder because `ListT` is not only deprecated, but also removed. See https://github.com/tweag/monad-bayes/issues/245. Probably, https://github.com/tweag/monad-bayes/pull/253 must be merged, or we need to provide some kind of `ListT` in the library.

turion:
Completed in https://github.com/tweag/monad-bayes/pull/267

PR #329, opened 2024-01-03
Bump mtl -> 2.3, transformers -> 0.6

turion:
This is the minimal change necessary to support the newer versions of `mtl` and `transformers`, both needed for #294. This doesn't solve #245 yet because it just copies the old `ListT` implementation, which is still invalid.


reubenharry (approved):
Seems good!

PR #342, opened 2024-05-13
Version bump to 1.3.0.1

turion:


turion:
CC @reubenharry

turion:
I need to bump the version so I can release a new version on hackage which includes the newer GHC 9.8 version bump

idontgetoutmuch (approved):

PR #379, opened 2025-10-07
Update brick & vty version bounds

turion:


turion:
This is unlikely to work without either a lot of work or #352

PR #381, opened 2025-10-07
Test 9.12 on CI

turion:
* [ ] #352 is needed to actually bring in this version of GHC

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -45,7 +45,7 @@ jobs:
     needs: lint
     strategy:
       matrix:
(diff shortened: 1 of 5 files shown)


turion:
Superseded by #388, which tests GHC 9.12 along with the flake.lock update this branch was blocked on (#352).

Two notes from redoing this on current nixpkgs:

- The `ghcs` list in `flake.nix` here never had `ghc912` added, only the CI matrix did, which is why `build-all-ghcs (ghc912)` failed on this branch: `monad-bayes-per-ghc.ghc912` did not exist.
- The `statistics` jailbreak turned out not to be needed anymore. All six GHCs build and pass tests without it.

The `time` bound relaxation and dropping 9.0/9.2 both carried over.

PR #388, opened 2026-07-29
Unblock CI: update flake.lock, test GHC 9.4-9.12, replace jupyenv

turion:
Supersedes #352 and #381.

`lint` was failing on #352, and since every other job `needs: lint`, the whole
matrix was skipped. Behind that were four independent problems:

1. **ormolu.** The ormolu version in current nixpkgs formats one list
   comprehension guard in `Control.Monad.Bayes.Integrator` differently.
2. **GHC versions.** GHC 9.0 and 9.2 are gone from nixpkgs, so `build-all-ghcs`
   could not work at all. Now testing 9.4 to 9.12. GHC 9.12 ships `time` 1.14,
   hence the wider upper bound; `base >= 4.17` reflects the lowest GHC we test.
3. **jupyenv.** `tweag/jupyenv` has been unmaintained since Dec 2024 and pulled
   its own years-old nixpkgs into our lock. It no longer even *evaluates* — its
   nixpkgs marks `ihaskell-hvega` as broken. Replaced with nixpkgs' own
   `ihaskell` plus `jupyter-kernel.create` for the kernelspec. This takes
   `flake.lock` from ~50 nodes to **7**.
4. **Notebook HTML.** nbconvert 7.17 emits a different template than whatever
   produced the committed files, so the notebooks job would fail on the diff.

Along the way, two long-standing bugs: the notebooks CI step and the README
both used `nix develop`, i.e. the devShell *without* Jupyter in it. Neither
could ever have worked. Both now use `.#default-jupyter`.

### Verified locally

- `nix build .#pre-commit` passes.
- All six GHCs build, and the test suite runs on each: 45 examples, 0 failures.
- `nix develop` on `monad-bayes` and on each `monad-bayes-per-ghc.ghc*`.
- `./regenerate_notebooks.sh` runs and produces the committed HTML.
- The kernel really works, not just builds — executing a notebook cell with
  `enumerator $ do { b <- bernoulli 0.7; return b }` yields
  `[(True,0.7),(False,0.30000000000000004)]`.

macOS is the one thing I could not check locally; that is on CI here.

### Overrides that changed

- `vty` override **dropped** — nixpkgs ships 6.4 itself now.
- `statistics` jailbreak from #381 turned out **not** to be needed.
- `brick` is still pinned: nixpkgs is on 2.9, the cabal file wants `^>=2.10`.

### Known pre-existing issues, deliberately not touched here

- `nix flake check` fails because `packages.monad-bayes-per-ghc` is an attrset
  rather than a derivation. CI does not run it, and that nested attribute is
  exactly what the `build-all-ghcs` matrix builds, so changing the shape would
  break CI.
- The notebooks declare kernelspec `ihaskell_nixpkgs` (`Diagrams.ipynb` uses
  `custom-haskell`), which matched neither the old jupyenv kernel nor the new
  one. This only affects opening them interactively; `nbconvert` without
  `--execute` does not resolve a kernel.
- `config.allowBroken = true` is no longer needed on jupyenv's account, but I
  left it in rather than risk surfacing unrelated breakage in this PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


turion:
@qnikst Do you have the rights to configure the repository? I removed the [x86_64-darwin](https://nixos.org/manual/nixpkgs/stable/release-notes) job, but the branch protection rules still expect the job.

qnikst:
@turion fixed that! Will review changes ASAP.

turion:
If you want, we can have a discussion or contributing guideline about LLM assistance somewhere

qnikst (approved):
Looks good to me!

qnikst:
@turion wrt a discussion and contributing guideline I've sent you a message in Tweag discord channel together with an example of the policy that I find effective and fair. I think it may be more productive to discuss that in a more live approach compared to GitHub. But I'm open to discussion on the other media as well!

turion:
I have a few minor additions

qnikst (approved):
Thanks! And thanks for the very nicely split commits with exhaustive messages!

qnikst (approved):

Issue #368, opened 2025-04-27
Tests fail on aarch64


ysangkok:
**Describe the bug**
After NixOS upgraded to monad-bytes-1.3.0.4, tests started failing.

**To Reproduce**
1. Use aarch64 (e.g. Mac M1)
2. Run monad-bayes test suite

**Expected behavior**
Tests should not fail.

**Environment**
 - OS name + version: nixpkgs with [Stackage LTS 23](https://github.com/NixOS/nixpkgs/pull/371032) merged in (i.e. staging branch)
 - Version of the code: monad-bayes-1.3.0.4

**Additional context**

Test run:
* https://hydra.nixos.org/build/295732332

I am posting this because I've gotten the impression that Tweag cares about NixOS, because there are many posts on the blog: https://www.tweag.io/group/nix/

turion:
Thanks for reporting! I believe this is a known, but not well-documented issue (https://github.com/tweag/monad-bayes/blob/bed154b1592e1d78dc51fefb87a0bd2e498a21fc/flake.nix#L48, https://github.com/tweag/monad-bayes/blob/bed154b1592e1d78dc51fefb87a0bd2e498a21fc/.github/workflows/nix.yml#L40). Due to machine-precision issues, floating point number calculations are slightly different on aarch64 than on x86_64. One could work around this by relaxing the equality check in the relevant tests. On the `monad-bayes` CI, we simply don't test aarch64.

> I am posting this because I've gotten the impression that Tweag cares about NixOS, because there are many posts on the blog: https://www.tweag.io/group/nix/

I guess Tweag cares about Nix and employs many Nix developers. But `monad-bayes` is currently not developed much.

My recommendation is to disable tests on aarch64 for now, or find someone with a Mac who wants to fix the tests.

PR #389, opened 2026-07-29
Build aarch64-linux in CI, pin down the fixture failures, and clean up the flake

turion:
Stacked on #388 — review that one first, this PR's base is `flake-update-fixes`.

## Build aarch64-linux in CI, and find out why the fixture tests fail

The Darwin test skip from #256 was written as "anything that isn't `x86_64-linux`":

```nix
modifier = drv:
  if system == "x86_64-linux"
  then drv
  else pkgs.haskell.lib.dontCheck drv;
```

so it silently disabled the test suite on `aarch64-linux` as well. That matters,
because it is exactly the configuration that tells us *why* the fixtures disagree
on Darwin. This PR adds an `aarch64-linux` job to CI on a `ubuntu-24.04-arm`
runner (free and unlimited on public repos) and runs the tests there.

### The hypothesis

The fixture tests in `test/TestSSMFixtures.hs` and `test/TestBenchmarks.hs` compare
`show`n `Double`s against committed fixtures, so a single ulp anywhere fails them.
Of everything in the sampling path:

* `mwc-random`'s core is integer arithmetic — bit-identical everywhere.
* `+ - * /` and `sqrt` are correctly rounded per IEEE 754 — bit-identical everywhere.
* `math-functions`' special functions are pure-Haskell rational/Chebyshev
  approximations — bit-identical everywhere.
* `log`, `exp`, `log1p`, `expm1` and `**` are foreign calls into the platform's
  libm, and IEEE 754 does *not* require these to be correctly rounded.

So the hypothesis was that the divergence is glibc (which the fixtures were
generated against) versus Apple's libm, not Intel versus ARM. `aarch64-linux`
holds the libm fixed and varies the architecture, which discriminates between
the two: if it passes, the cause is the libm; if it fails, the cause is
architectural and the reasoning above is wrong.

### The answer: it is the architecture

It fails. [Run 30532566710](https://github.com/tweag/monad-bayes/actions/runs/30532566710/job/90838226558?pr=389),
`aarch64-linux`, glibc, 45 examples, 5 failures:

```
1) Benchmarks       LR10-SMC
2) TestSSMFixtures  SSM-RMSMC
3) TestSSMFixtures  SSM-RMSMCDynamic
4) TestSSMFixtures  SSM-RMSMCBasic
5) TestSSMFixtures  SSM-SMC2
```

and the deviations are exactly one ulp in a single element, e.g. in `SSM-RMSMCBasic`:

```
expected: -362.48592096242163
 but got: -362.4859209624216
```

So Apple's libm is not the cause, glibc on aarch64 does the same thing, and no
amount of changing the Darwin architecture would have helped. Correspondingly
this PR now:

* keeps the `aarch64-linux` job, because it builds the platform nixpkgs ships
  aarch64 from (#368) and it is where a fix can be verified without a Mac;
* skips the test suite on **aarch64** rather than on **Darwin**, which is what
  the condition always meant to say;
* records the finding in the comment, instead of the wrong glibc-vs-Apple guess.

The "Find out what is different, and use a type with system-independent
precision" checkbox from #256 is still unchecked, but the first half of it is
now answered.

### Follow-up

Making the tests actually pass on aarch64 means not comparing `show`n `Double`s
any more, as discussed below with @qnikst: serialise the fixtures as JSON and
compare numerically with an epsilon, or wrap the result type in a newtype whose
`Show` rounds to a fixed number of digits. That is a separate PR; #368 tracks it.

## Cleanup

Second commit, no behaviour change intended:

* `config.allowBroken = true` is no longer needed now that we build Jupyter from
  nixpkgs instead of jupyenv. Verified by evaluating every output CI builds
  without it: `monad-bayes`, all five `monad-bayes-per-ghc` entries,
  `jupyterEnvironment`, `pre-commit` and the dev shells.
* `microstache` no longer needs `doJailbreak` on GHC 9.10 or 9.12 — both build
  unpatched straight from the binary cache.
* `defaultPackage`/`devShell` were there for Nix < 2.8, which is over three years
  old. `flake-compat` prefers `packages.default`/`devShells.default` over them
  anyway, so `default.nix` and `shell.nix` keep working.
* `nix develop .#ghc94` was broken: `haskell-language-server` dropped GHC 9.4 in
  2.12.0.0 and nixpkgs `throw`s on it rather than marking it broken. Now only
  added for GHC >= 9.6.
* `"\."` in a Nix string is an ill-defined escape and warns on every evaluation.
* `devShellFor` never used its `ghcVersion` argument.
* The `notebooks` job had a one-entry matrix plus a commented-out `x86_64-darwin`
  entry for a system nixpkgs no longer supports.

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -27,6 +27,12 @@ jobs:
         include:
           - os: ubuntu-latest
             system: x86_64-linux
+          # aarch64-linux told us that the fixture test failures are about the
+          # architecture and not about Apple's libm, see the `modifier` in
+          # flake.nix. Keep building it so the platform nixpkgs ships from stays
+          # green, and so we notice when the tests can be turned back on there.
+          - os: ubuntu-24.04-arm
+            system: aarch64-linux
           # The macos-latest runners are arm64, and nixpkgs 26.11 dropped
           # x86_64-darwin altogether.
           - os: macos-latest
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -78,19 +78,13 @@ jobs:
       - name: Development environment (package only)
         run: nix --print-build-logs develop .#packages.${{ matrix.system }}.monad-bayes-per-ghc.${{ matrix.ghc }} --command echo Ready
 
+  # Only on x86_64-linux: this job asserts that the committed *.html matches what
+  # regenerating the notebooks produces, and that output is platform dependent.
   notebooks:
     needs:
       - build
       - build-all-ghcs
-    strategy:
-      matrix:
-        include:
-          - os: ubuntu-latest
-            system: x86_64-linux
-          # The Jupyter kernel is only tested on Linux so far
-          # - os: macos-latest
-          #   system: x86_64-darwin
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v7
       - name: Setup Nix Environment
--- a/flake.nix
+++ b/flake.nix
@@ -35,11 +35,11 @@
     } @ inputs:
     flake-utils.lib.eachSystem
       [
-        # Tier 1 - Tested in CI
+        # All of these are built in CI, see .github/workflows/nix.yml. The test
+        # suite only runs on x86_64 though, see the `modifier` below.
         flake-utils.lib.system.x86_64-linux
-        flake-utils.lib.system.aarch64-darwin
-        # Tier 2 - Not tested in CI (at least for now)
         flake-utils.lib.system.aarch64-linux
+        flake-utils.lib.system.aarch64-darwin
         # Note: no x86_64-darwin. nixpkgs 26.11 dropped support for it, and
         # importing nixpkgs for that system now throws rather than merely warns.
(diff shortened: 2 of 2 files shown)


qnikst:
@turion I've tried to debug the problems with the arithmetic precision locally on aarch64 (though Darwin, but I expect that the reason for the failure is the same) and have not found the easy way to bypass this limitation, the way how comparison is done is not vary compatible with the non-precise comparison. The changes that I saw was very substantial to propose.

Do you have an idea what we can do there?

turion:
I think the right solution is to just accept minor rounding errors and relax the test suite such that it accepts deviations below an `epsilon = 0.00001` or so.

qnikst:
I agree with that but if I interpret code correctly it does a comparison of the string representation. And to apply an approximate comparison we need to re-interpret the data. While I think it’s a reasonable approach I find myself uneasy to suggest that and maybe there are better way forward. E.g as it is abstract over the result type we can have a newtype wrapper that will round down number to a certain digit on show.

turion:
Yes, for example. I'd favour ex-/importing it as JSON, though.

qnikst (approved):

qnikst:
All updates looks fine to me!

PR #390, opened 2026-07-30
Regenerate the tutorial notebooks too, so the docs stop drifting

turion:
Follow-up to #388. Targets `flake-update-fixes`, so GitHub retargets this to
`master` once #388 lands.

`regenerate_notebooks.sh` only ever converted `notebooks/examples`, so the seven
pages generated from `notebooks/tutorials` kept whatever template produced them
years ago. Now that #388 regenerates the examples with nbconvert 7.17, the site
splits exactly along its own navigation:

| Nav page | Template before this PR |
| --- | --- |
| `examples.md` — Histogram, Lenses, Diagrams, Parsing, Streaming, Ising, ClassicalPhysics | nbconvert 7.17 |
| `tutorials.md` — Introduction, Sampling, Bayesian, MCMC, SMC, Lazy, AdvancedSampling | older JupyterLab template |

These are standalone documents that mkdocs copies verbatim, each carrying its
own stylesheet, so the two sets really do render differently: 77 of ~300 `--jp-*`
custom properties differ, including `--jp-content-font-family` and
`--jp-ui-font-family`, which now lead with `system-ui`. Code cells moved from
CodeMirror 5's `.CodeMirror` to CodeMirror 6's `.cm-editor`.

The more lasting problem is the CI gate. The `notebooks` job runs
`regenerate_notebooks.sh` and then `git diff --exit-code`; a file the script
never touches is checked-in build output that nothing verifies. It could already
have been stale against its `.ipynb` and CI would not have noticed. Widening the
glob puts the tutorials under the same gate.

### Verified

- **No content changes.** Code cells, markdown cells, outputs and embedded
  images all match the previously committed files exactly, for all seven.
- **Reproducible.** Running the script twice leaves the tree clean, so the CI
  diff gate is stable.
- Each page shrinks by roughly 300 kB, which is the smaller stylesheet 7.17
  embeds:

  ```
  AdvancedSampling 930→629 kB   Bayesian 842→544 kB   Introduction 822→523 kB
  Lazy 643→344 kB   MCMC 899→599 kB   Sampling 846→546 kB   SMC 795→495 kB
  ```

### Left alone

`docs/docs/notebooks/Functional_PPLs.html` (7.2 MB) and `RealTimeInference.html`
(still on nbconvert 5.x's classic template, loading Bootstrap 3.3.7 and jQuery
from cdnjs) have **no source `.ipynb` anywhere in the repository**, and nothing
in `docs/docs/*.md` links to them. They cannot be regenerated, so they need a
separate decision — restore the sources, or delete the files. They are the only
two pages that still differ from the rest after this PR.


qnikst (approved):

qnikst:
Thanks a lot for looking into that!

